PR #7, opened 2020-01-28
(interpreter) Add "quit" command to REPL

perlun:
(I realize we should move the REPL to a project of its own, perhaps `Perlang.REPL` or something. It feels silly to have it inside the `Perlang.Interpreter` project.)


perlun:
> (I realize we should move the REPL to a project of its own, perhaps `Perlang.REPL` or something. It feels silly to have it inside the `Perlang.Interpreter` project.)

For reference, fixed in #13.

Issue #12, opened 2020-01-30
Fix broken for loops


perlun:
Certain `for` loops are currently broken:

```
> for (var i = 0; i < 10; i = i + 1) print i;
Unhandled exception. System.Collections.Generic.KeyNotFoundException: The given key 'i' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Perlang.Interpreter.PerlangEnvironment.GetAt(Int32 distance, String name) in /home/per/git/perlang/Perlang.Interpreter/PerlangEnvironment.cs:line 26
   at Perlang.Interpreter.PerlangInterpreter.LookUpVariable(Token name, Expr expr) in /home/per/git/perlang/Perlang.Interpreter/PerlangInterpreter.cs:line 218
   at Perlang.Interpreter.PerlangInterpreter.VisitVariableExpr(Variable expr) in /home/per/git/perlang/Perlang.Interpreter/PerlangInterpreter.cs:line 211
   at Perlang.Expr.Variable.Accept[TR](IVisitor`1 visitor) in /home/per/git/perlang/Perlang.Common/Expr.cs:line 161
   at Perlang.Interpreter.PerlangInterpreter.Evaluate(Expr expr) in /home/per/git/perlang/Perlang.Interpreter/PerlangInterpreter.cs:line 294
   at Perlang.Interpreter.PerlangInterpreter.VisitPrintStmt(Print stmt) in /home/per/git/perlang/Perlang.Interpreter/PerlangInterpreter.cs:line 361
   at Perlang.Stmt.Print.Accept[TR](IVisitor`1 visitor) in /home/per/git/perlang/Perlang.Common/Stmt.cs:line 98
   at Perlang.Interpreter.PerlangInterpreter.Execute(Stmt stmt) in /home/per/git/perlang/Perlang.Interpreter/PerlangInterpreter.cs:line 299
   at Perlang.Interpreter.PerlangInterpreter.ExecuteBlock(IEnumerable`1 statements, PerlangEnvironment blockEnvironment) in /home/per/git/perlang/Perlang.Interpreter/PerlangInterpreter.cs:line 317
   at Perlang.Interpreter.PerlangInterpreter.VisitBlockStmt(Block stmt) in /home/per/git/perlang/Perlang.Interpreter/PerlangInterpreter.cs:line 328
[...]

PR #15, opened 2020-01-31
(tests) Add `for`-tests

perlun:
This also included some changes to the interpreter to be able to capture the output from `print` statements. This is really nice, since it makes adding the full Lox test suite much simpler (which is something that I will probably do, a small chunk at a time, for the relevant parts).

### DIFF
--- a/Perlang.Tests/For/Syntax.cs
+++ b/Perlang.Tests/For/Syntax.cs
@@ -1,0 +1,129 @@
+using Xunit;
+using static Perlang.Tests.EvalHelper;
+
+namespace Perlang.Tests.For
+{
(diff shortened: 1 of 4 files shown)


Perlang.Tests/For/Syntax.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            Assert.Equal(new[]
+            {
+                "1",
+                "2",
+                "3"
+            }, output);
+        }
+
+        [Fact(Skip = "Broken because of https://github.com/perlun/perlang/issues/12")]
```
perlun: That is, #12 for reference.

PR #17, opened 2020-02-02
(all) Add support for postfix increment and decrement

perlun:
~This is not ready yet, will complete another day.~ Reasonably complete now.

Things currently missing:

- [x] Support for decrement also, not just increment
- [ ] _Ensure that we put this at a reasonable level of precedence. (The way I did it now was just a slightly lame approximation)_

  Maybe we can live with this as it is now after all. Right now I put it between "assignment" and "logical or". I would love to detect invalid usages such as `foo(c++)` which I consider very bad practice, but we can perhaps live without that at this stage. _Note_: even though these are postfix operators at the moment, they work like _prefix_ operators do in other languages. I.e. `var c = 0; some_function(c);` will use `1` as the value of the the parameter passed to the function.

  I tried to document this now in the syntax grammar, which is btw again heavily inspired by the Lox specification: #18.
- [x] Properly handle cases when the user tries to do things like `var a = "foo"; a++;` and similar folly. This is clearly an area where a static type checker would make things immensely much better; then interpreter/compiler in the future will then have to take much less precautions in ensuring that things are of the expected types all the time.

### DIFF
--- a/Perlang.Common/Expr.cs
+++ b/Perlang.Common/Expr.cs
@@ -18,7 +18,8 @@ namespace Perlang
             TR VisitGroupingExpr(Grouping expr);
             TR VisitLiteralExpr(Literal expr);
             TR VisitLogicalExpr(Logical expr);
-            TR VisitUnaryExpr(Unary expr);
+            TR VisitUnaryPrefixExpr(UnaryPrefix expr);
+            TR VisitUnaryPostfixExpr(UnaryPostfix expr);
             TR VisitVariableExpr(Variable expr);
         }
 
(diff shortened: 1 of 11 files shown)


perlun:
Making process, but there are still some things that are broken - the failing tests indicate this. There is _something_ causing the resolution of local variables to fail, which is quite odd  given that _prefix_ operators work flawlessly.

I tried moving the unary postfix operators into a separate method now (instead of just tucking it into the `Assignment()` method), but it doesn't change a thing. It feels like there some oddness in the variable resolver going on; I guess I just have to continue digging...

perlun:
> Making process, but there are still some things that are broken - the failing tests indicate this. There is _something_ causing the resolution of local variables to fail, which is quite odd given that _prefix_ operators work flawlessly.

The thing missing was the `ResolveLocal()` call in `VisitUnaryPostfixExpr()`; this is necessary to be able to resolve variables and other locals. Isn't it interesting how things sometimes are much easier if you leave it for a while and revisit the problem the next day? :slightly_smiling_face:

perlun:
> * _Note_: even though these are postfix operators at the moment, they work like _prefix_ operators do in other languages. I.e. `var c = 0; some_function(c++);` will use `1` as the value of the the parameter passed to the function.

This was subsequently revisited in #145. Using different semantics than other languages here was simply discovered to be a pretty bad idea.

PR #20, opened 2020-02-11
(all) Add CI & config for publishing packages on each `master` commit

perlun:
The idea is to build & publish builds as plain `.tar.gz` archives for each of the following targets on each push to the `master` branch:

- `linux-x64`
- `osx-x64`
- `win-x64`

The builds are available here: https://dl.bintray.com/perlang/builds/

We will also (perhaps at a slightly later stage) publish proper `.deb` packages for Ubuntu 16.04+ and Debian 9+, since this makes it much easier to always get the latest updates to the Perlang tooling.

### DIFF
--- a/Perlang.ConsoleApp/Perlang.ConsoleApp.csproj
+++ b/Perlang.ConsoleApp/Perlang.ConsoleApp.csproj
@@ -3,6 +3,8 @@
     <PropertyGroup>
         <OutputType>Exe</OutputType>
         <TargetFramework>netcoreapp3.1</TargetFramework>
+        <PublishTrimmed>true</PublishTrimmed>
+        <AssemblyName>perlang</AssemblyName>
     </PropertyGroup>
 
     <ItemGroup>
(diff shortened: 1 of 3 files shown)


Perlang.ConsoleApp/Perlang.ConsoleApp.csproj
```diff
@@ -3,6 +3,8 @@
     <PropertyGroup>
         <OutputType>Exe</OutputType>
         <TargetFramework>netcoreapp3.1</TargetFramework>
+        <PublishTrimmed>true</PublishTrimmed>
```
perlun: This uses the Mono Linker to make the output smaller. More info: https://github.com/mono/linker/blob/master/src/ILLink.Tasks/README.md

It would also be useful to look into some of the suggestions given here, which can make us build a more convenient "fat" executable. But we need to ensure that we don't loose too much in runtime performance by doing so. (in fact, the R2R suggestions given there are likely to _improve_ startup performance) https://www.hanselman.com/blog/MakingATinyNETCore30EntirelySelfcontainedSingleExecutable.aspx

PR #25, opened 2020-02-18
(all) Add base64_encode function

perlun:
`base64_decode` will follow shortly.

The unrelated `RuntimeError` -> `Exception` widening of runtime errors is needed so we can properly handle exceptions in 3rd party code, custom (native) functions etc. Thinking more about it, perhaps an even better way to handle this would be to let the interpreter method handling the callable catch `Exception` and propagate it as a `RuntimeError` instead. The big advantage of this would be that we can then get a proper line location included in the `RuntimeError`, which is much more helpful to users when debugging problems in their programs.

Holding back on merging this until I've investigated that route a bit.

### DIFF
--- a/Perlang.Interpreter/PerlangInterpreter.cs
+++ b/Perlang.Interpreter/PerlangInterpreter.cs
@@ -551,11 +551,25 @@ namespace Perlang.Interpreter
 
             if (arguments.Count != function.Arity())
             {
-                throw new RuntimeError(expr.Paren, "Expected " + function.Arity() + " arguments but got " +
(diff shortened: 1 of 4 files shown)


perlun:
> Holding back on merging this until I've investigated that route a bit.

This turned out to be clearly doable, so PR has been adjusted. Errors thrown because of invalid types are now much cleaner:

```
$ dotnet run
Perlang Interactive REPL Console (0.1.0)
> base64_encode(42)
[line 1] base64_encode: Unable to cast object of type 'System.Double' to type 'System.String'.
```

PR #32, opened 2020-02-24
(many) Add git revision into generated binaries

perlun:
~This is still WIP, will finalize this another day. The idea is to make it more apparent (in the REPL) what git version a given build represents.~ Should be complete now.


Perlang.Common/Perlang.Common.csproj
```diff
@@ -1,9 +1,19 @@
...
     <PropertyGroup>
         <TargetFramework>netcoreapp3.1</TargetFramework>
         <RootNamespace>Perlang</RootNamespace>
         <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     </PropertyGroup>
 
+    <!-- Must come before the $(ProjectDir) can be referenced -->
+    <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
+    <PropertyGroup>
+        <PreBuildEvent>$(SolutionDir)/scripts/update_common_constants.rb $(SolutionDir)</PreBuildEvent>
```
perlun: In theory, it should be enough to run this. But in reality, it didn't; the CI builds were failing. I can reproduce it by doing the following locally:

```shell
$ rm Perlang.Common/CommonConstants.Generated.cs 
$ dotnet build
Microsoft (R) Build Engine version 16.4.0+e901037fe for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Restore completed in 36.47 ms for /home/per/git/perlang/Perlang.Common/Perlang.Common.csproj.
  Restore completed in 39.6 ms for /home/per/git/perlang/Perlang.Parser/Perlang.Parser.csproj.
  Restore completed in 39.59 ms for /home/per/git/perlang/Perlang.Interpreter/Perlang.Interpreter.csproj.
  Restore completed in 39.6 ms for /home/per/git/perlang/Perlang.Stdlib/Perlang.Stdlib.csproj.
  Restore completed in 43.51 ms for /home/per/git/perlang/Perlang.ConsoleApp/Perlang.ConsoleApp.csproj.
  Restore completed in 99.7 ms for /home/per/git/perlang/Perlang.Tests/Perlang.Tests.csproj.
  /home/per/git/perlang/Perlang.Common/CommonConstants.Generated.cs updated
Properties/AssemblyInfo.cs(5,89): error CS0117: 'CommonConstants' does not contain a definition for 'GitVersion' [/home/per/git/perlang/Perlang.Common/Perlang.Common.csproj]

Build FAILED.

Properties/AssemblyInfo.cs(5,89): error CS0117: 'CommonConstants' does not contain a definition for 'GitVersion' [/home/per/git/perlang/Perlang.Common/Perlang.Common.csproj]
    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:04.17
```

I _think_ this is because MSBuild reads the files first, then runs the pre-build event(s), then builds the project. If the pre-build event generates a new file, it simply won't work.

As suggested by https://stackoverflow.com/questions/20323012/visual-studio-c-sharp-does-not-build-files-that-were-change-by-pre-build-event, I tried tweaking the `UseHostCompilerIfAvailable` setting but to no avail.

perlun: I chose a different way, since I couldn't get the above working in any reasonable way. By moving the generation of this file out to a command that gets executed _before_ `dotnet build`, we eliminate this problem entirely.

Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+clean:
+	dotnet clean
+	rm -f Perlang.Common/CommonConstants.Generated.cs
+
+install: all
+	./scripts/local_install_linux.sh
+
+run:
+	cd Perlang.ConsoleApp && dotnet run
```
perlun: `make` is clearly not the best tool in the world for handling building of software systems, but its ubiquity and bookshelves full of documentation make it useful. I've used Rake (in a fairly sophisticated manner) in other projects and I'm not so fond of it since it tends to become rather complex over time. That's not to say that Makefiles are always _easy_ but... let's at least try to keep them this way in this case.

As can be seen, I also added a few other useful targets while I was at it:

- `clean`
- `install`
- `run`

PR #42, opened 2020-05-14
    (many) Store the TypeReference in the Resolver

perlun:
This is a "preparational commit", aimed at making support for static typing simpler. Will hold off on merging this a while, until I see that it's actually useful in implementing that PR. (I have a WIP PR for that, but it will probably take a couple of weeks or so before it's in any shape near mergeability...)

### DIFF
--- a/Perlang.Common/TypeReference.cs
+++ b/Perlang.Common/TypeReference.cs
@@ -1,0 +1,77 @@
+using System;
+
+namespace Perlang
+{
+    /// <summary>
+    /// A type reference is a (potentially unresolved) reference to a Perlang type.
+    /// </summary>
+    public class TypeReference
+    {
+        /// <summary>
+        /// A "None" type reference, indicating that no type information is available. Note that this is different
+        /// from an unresolved type reference, which is the state of a TypeReference before type inference has taken
+        /// place.
+        /// </summary>
+        public static TypeReference None { get; } = new TypeReference(null);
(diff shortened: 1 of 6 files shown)


Perlang.Common/TypeReference.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                return IsResolved ? $"Explicit: {ClrType}" : $"Explicit: {TypeSpecifier}";
+            }
+            else
+            {
+                return IsResolved ? $"Inferred: {ClrType}" : "Inferred, not yet resolved";
+            }
+        }
+    }
+}
```
perlun, line 77: Note: this class isn't really used yet, but it was easiest to just include it as-is in this PR. It will be used in one of the subsequent steps, where we add support for type specifiers.

Perlang.Interpreter/Resolver.cs
```diff
@@ -280,7 +292,8 @@ public VoidObject VisitVarStmt(Stmt.Var stmt)
                 Resolve(stmt.Initializer);
             }
 
-            Define(stmt.Name);
+            // TODO: Get the type reference from the initializer instead.
+            Define(stmt.Name, TypeReference.None);
```
perlun, line 296: (Again, the necessary preconditions for putting a real `TypeReference` value in place here isn't really there yet.)

Issue #43, opened 2020-05-16
Support optional typing and type inference


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/43. The GitHub page you are currently reading will not contain the latest information on this issue.

---

### Background

Perlang is currently a _dynamically_ typed language, which has its explanation in that it has its roots as being a friendly _fork_ of [the Lox language](https://github.com/munificent/craftinginterpreters), as described in @munificent's great programming-language-tutorial ["Crafting Interpreters"](http://craftinginterpreters.com/). This has some advantages in programmer ergonomics (no need to spell out types), and it also greatly simplifies the work of implementing the interpreter.

_However_, as many of us have already experienced, this comes at a price. Purely dynamically typed languages can get frustrating to work with, especially as the project grows in complexity:

* **Harder to refactor code** (especially if your code lacks proper unit tests). With languages like Ruby, you basically need 100% test coverage for all `if`-branches in your program, for all the permutation of types being passed as variables to _all_ methods in your program. Otherwise, you are in risk of breaking your program as you refactor it.

* **Inferior performance**. If the language runtime/compiler knows the types of variables as they are flowing through your program, it can perform runtime or compile-time optimizations to increase the performance of the program. This difference is often significant; I am yet to see a dynamically typed language with performance near statically typed languages like C, C++ and Rust. Even byte-code-interpreted/JIT:ed languages like C# and Java are typically faster than the dynamically typed languages, and the static typing is one part of this. See for yourself on example of a benchmark that illustrates this: https://benchmarksgame-team.pages.debian.net/benchmarksgame/performance/binarytrees.html (The Common Lisp program is doing remarkably well there, despite Common Lisp being a dynamically typed language)

* **More frustrating to write code in**. This is of course highly subjective, many would argue to the contrary. For me personally, I find it annoying to not be able to "lock down" a method saying that _"parameter foo can only be a string, parameter bar can only be a bool"_ and so on. I tend to have that information in my head when writing the code, and being "forced" (by languages like Ruby) to throw away this information can feel frustrating, especially when you know that with some static analysis, that information could be helpful for the programming language to detect at a very early stage whether the program is "broken" or not (at least not from a "mixing types" perspective). In this sense, this is not just a subjective point. **Better adherence to the "fail-fast" paradigm"** is an objective point we could highlight here.

[...]

PR #44, opened 2020-05-16
(tests) Add comparison operator tests

perlun:
This was added as a side effect of working on #43. It's all based on the great prior art done by @munificent in https://github.com/munificent/craftinginterpreters/blob/master/test/operator/comparison.lox. (Thank you, Bob! :heart:)

### DIFF
--- a/Perlang.Tests/Operator/Comparison.cs
+++ b/Perlang.Tests/Operator/Comparison.cs
@@ -1,0 +1,268 @@
+using System.Linq;
+using Xunit;
+using static Perlang.Tests.EvalHelper;
+
+namespace Perlang.Tests.Operator
+{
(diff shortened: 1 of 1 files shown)


Perlang.Tests/Operator/Comparison.cs
```diff
@@ -0,0 +1,268 @@
...
+
+            string output = EvalReturningOutput(source).SingleOrDefault();
+
+            Assert.Equal("True", output);
+        }
+
+        //
+        // Special tests for 0 and -0, ensuring that positive and negative zero compares as the same value.
+        //
+        [Fact(Skip = "Fails with 'Operand must be a number' error")]
```
perlun, line 172: I discovered when implementing these tests that this is broken. Issue filed about it: https://github.com/perlun/perlang/issues/45

Issue #46, opened 2020-05-17
Incorporate tests from Lox


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/46. The GitHub page you are currently reading will not contain the latest information on this issue.

---

There are a large number of tests at https://github.com/munificent/craftinginterpreters/tree/master/test. We implement some of them, but far from all.

We should work towards incorporating most/all of these, since they are useful for us also.

## Status

- [x] assignment - supported as of #130
- [ ] benchmark
- [ ] block 
- [ ] bool
- [ ] call
- [ ] class
- [ ] closure
- [x] comments - supported as of #141
- [ ] constructor
- [ ] empty_file.lox
- [ ] expressions
- [ ] field
- [ ] for - partially supported (#15)
- [ ] function - partially supported (#48)
- [ ] if
- [ ] inheritance
- [ ] limit
- [ ] logical_operator - partially supported (#47)
- [ ] method
- [ ] nil
- [x] number - supported as of #71
- [ ] operator - partially supported (#44, #63, #151)
- [x] precedence.lox - supported as of #4
- [ ] print
- [ ] regression
- [x] return - supported as of #51
- [ ] scanning
- [ ] string
- [ ] super
- [ ] this
- [ ] unexpected_character.lox
- [x] variable - supported as of #33
- [ ] while

PR #48, opened 2020-05-19
(tests) #46 Add function tests

perlun:
Note: this does not include all tests, but at least 13 full tests, all of which run successfully with the current `master` codebase.

### DIFF
--- a/Perlang.Tests/Blocks.cs
+++ b/Perlang.Tests/Blocks.cs
@@ -7,7 +7,7 @@ namespace Perlang.Tests
     public class Blocks
     {
         [Fact]
-        public void calling_an_undefined_function_inside_a_block_throws_expected_exception()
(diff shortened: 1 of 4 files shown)


Perlang.Tests/Blocks.cs
```diff
@@ -17,7 +17,7 @@ public void calling_an_undefined_function_inside_a_block_throws_expected_excepti
         }
 
         [Fact]
-        public void referring_to_an_undefined_variable_inside_a_block_throws_expected_exception()
+        void referring_to_an_undefined_variable_inside_a_block_throws_expected_exception()
```
perlun, line 20: Drive-by change. I think we could perhaps start dropping the `public` part for tests. It's not really needed and the test code is arguably a bit more readable without it (less noise when skimming through test classes).

PR #54, opened 2020-06-06
(all) #43 Implement support for optional typing

perlun:
This took a while :wink: but I'm quite happy about the overall result. Will wait a day or two before merging this, but all unit tests are now passing at least, and most WIP code should be cleaned up.

## Supported functionality as of this MR

- Types can be specified for local variables. They can also be inferred from their usage.

  ```perlang
  var a: int = 32;
  var a = 42; // inferred to int
  ```

- Variable types can also be inferred from e.g. calling functions, the result of binary operators etc:

  ```perlang
  fun f(): int {
    return 42;
  }

  var a = f();          // inferred to int
  var b = 32 * 42 * 52; // likewise
  ```

## Limitations

- Function parameters and return types are currently *mandatory*; they cannot be inferred from their usage (as in TypeScript). This is something we aim to fix within the scope of #43 in the longer run.

  ```perlang
  // The type specifiers below are currently mandatory
  fun f(s: string): int {
  }
  ```


Perlang.Common/Stmt.cs
```diff
@@ -4,6 +4,7 @@
 // Instead, change the ./scripts/generate_ast_classes.rb script that generated this code.
 //
```
perlun, line 5: Note: this is no longer the case. We should remove `scripts/generate_ast_classes.rb` now that we are grown out of it, and remove this comment.

perlun: Fixed, resolving.

PR #59, opened 2020-06-14
(docs) Update README.md, minor installer tweaks

perlun:


scripts/perlang-install
```diff
@@ -137,7 +137,7 @@ main() {
     if $ansi_escapes_are_valid; then
         # We presume that the likeliness of Unicode emojis working is higher in
         # case ANSI escape sequences are valid.
-        echo "🎉 \33[1mInstallation complete 🎉\33[0m"
+        printf "🎉 \33[1mInstallation complete 🎉\33[0m\n"
```
perlun, line 140: I noted that `echo` with escape statements doesn't work on `bash`, `\33[1m` is printed in plain instead of enabling bold/white text.

scripts/local_install_linux.sh
```diff
@@ -2,14 +2,14 @@
...
 
 set -e
 
-mkdir -p $HOME/.perlang/bin
+mkdir -p $HOME/.perlang/nightly/bin
 
 pushd Perlang.ConsoleApp
 dotnet publish -c Release -r linux-x64 --self-contained true /p:PublishReadyToRun=true
-cp -r bin/Release/netcoreapp3.1/linux-x64/publish/* $HOME/.perlang/bin
+cp -r bin/Release/netcoreapp3.1/linux-x64/publish/* $HOME/.perlang/nightly/bin
```
perlun, line 14: Script was changed to harmonize paths with the ones used by the installer.

Issue #39, opened 2020-05-07
Enable nullable references


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/39. The GitHub page you are currently reading will not contain the latest information on this issue.

---

Nullable references (or rather, _non-nullable_ references) is one of the cooler features in more recent versions of C# (more specifically, [C# 8.0](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8) or greater - .NET Core 3.0 and upwards) We have been a bit sloppy here, but we should enable these compiler flags to try and take advantage of this support right from start, especially since the Perlang language has strong ambitions regarding non-nullability in itself.

perlun:
Some of this sneaked in as part of #77 being merged. I found out that it's possible to enable nullability warnings for individual files ([like this](https://github.com/perlun/perlang/pull/77/files#diff-484e955e8fd3efee31d97f356f631d61R1)). This is incredibly convenient, since it means we can gradually start moving towards more and more nullability types until the whole codebase is converted to use it.

The newly created `Perlang.Tests` project also has nullability support enabled project-wide. `Perlang.Tests.Integration` could be a reasonable next project to enable it for.

perlun:
Not (fully) accomplished in 0.2.0, postponing to a later iteration.

PR #79, opened 2020-09-19
(many) Add support for native classes + calling static methods in native classes

perlun:
This was not trivial, since for some of the use cases (`Foo.to_string().to_string()`) I had to implement support for method overloading at the same time. Method overloading is such an anti-pattern, at least from a compiler/interpreter point of view; it definitely makes things more complex.

Then again, from a user point-of-view, it has its advantages in the polymorphic designs it enables. I _guess_ we will have to implement it for Perlang classes at some point. For now, we support it when interfacing with native .NET classes, which is a start. (Speaking about it, we have probably accomplished much of the complexity mentioned above already, so maybe, _maybe_ implementing it for Perlang classes would be quite "easy". Time will tell. :grin:)


Perlang.Common/Expr.cs
```diff
@@ -88,7 +90,7 @@ public class Call : Expr
             public Token Paren { get; }
             public List<Expr> Arguments { get; }
 
-            public string CalleeToString
+            public string? CalleeToString
```
perlun, line 93: Again, some #39-related changes sneaked into this MR.

Perlang.Common/Expr.cs
```diff
@@ -240,17 +242,21 @@ public override TR Accept<TR>(IVisitor<TR> visitor)
             }
 
             public override string ToString() =>
-                Name.Lexeme;
+                $"#<Identifier {Name.Lexeme}>";
```
perlun, line 245: This was not related to this change per se, but it looked weird when these expressions were being used in "key not found" exceptions. It makes more sense (it looked like it was a string being used as key).

PR #83, opened 2020-10-02
(ci) Auto-publish website

perlun:
The provided GitHub Action is now set up to publish updates whenever any relevant files are changed on the `master` branch. This should make it much more convenient to push changes to https://perlang.org than before; no more manual `rsync`ing from a developer laptop etc.

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - master
+    paths:
+      # TODO: Replace Perlang** with src/ once we have moved all the projects in there.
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-packages.yml
```diff
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - master
+    paths:
+      # TODO: Replace Perlang** with src/ once we have moved all the projects in there.
+      # TODO: https://github.com/perlun/perlang/issues/82
+      - Perlang**
```
perlun, line 10: (Drive-by-change, trying to avoid publishing packages unless some relevant file(s) have been changed.)

PR #86, opened 2020-10-04
(consoleapp) #60: Persist REPL state between commands

### LINKED ISSUE #60: Fix REPL state to persist between commands
Things like this is currently broken:

```
$ perlang
Perlang Interactive REPL Console (0.1.0+ea4f649)
> var s = "hej";
> print(s);
[line 1] Error at 's': Undefined variable 's'
```

perlun:
Closes #60


Perlang.Tests/ConsoleApp/ProgramTest.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    public class ProgramTest
+    {
+        private readonly Program subject;
+        private readonly List<string> output = new List<string>();
+
+        public ProgramTest()
+        {
+            subject = new Program(standardOutputHandler: s => output.Add(s), runtimeErrorHandler: e => throw e);
+        }
```
perlun, line 15: This class adds some rudimentary testing of the `Program`, i.e. the `perlang` entry point. We can expect more to be added in this are as we move along. :tada:

Perlang.Tests/ConsoleApp/ProgramTest.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            // This is definitely a bug. The second statement should cause an error, since a global variable with the
+            // same name already exists. For now, this test will remember us that these are the currently expected
+            // semantics.
+            subject.Run("var a = 42;");
+            subject.Run("var a = 44;");
+            subject.Run("print a;");
+
+            Assert.Equal(new List<string> {"42"}, output);
+        }
```
perlun, line 45: This is of course not how it _should_ be, but I'm too tired to have the energy to fix it properly right now. We should at least make a proper issue about fixing it sometime.

perlun: I managed to find and fix this problem, but now I have another one instead... Will write about it separately.

Perlang.Tests/ConsoleApp/ProgramTest.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        [Fact]
+        public void Run_state_does_not_persist_if_one_statement_is_invalid()
+        {
+            // TODO: This program should fail. Try to find out why.
+            subject.Run("var a = 42;");
+            subject.Run("var b = 43; x;");
+            subject.Run("print b");
+
+            Assert.Equal(new List<string> {"42"}, output);
```
perlun, line 42: This is incredibly weird. Here is the result from running this test:

```
Perlang.Tests.ConsoleApp.ProgramTest.Run_state_does_not_persist_if_one_statement_is_invalid

Xunit.Sdk.EqualException
Assert.Equal() Failure
Expected: List<String> ["42"]
Actual:   List<String> []
   at Perlang.Tests.ConsoleApp.ProgramTest.Run_state_does_not_persist_if_one_statement_is_invalid() in /home/per/git/perlang/Perlang.Tests/ConsoleApp/ProgramTest.cs:line 42
```

No error, no printing, no nothing. :thinking:

Perlang.sln
```diff
@@ -49,4 +53,13 @@ Global
...
 	EndGlobalSection
+	GlobalSection(NestedProjects) = preSolution
+		{B338D29B-A85D-4650-9BFF-FBFE55258C58} = {B924E83B-6AFE-4ABF-B79D-444F92582892}
+		{EC950B39-0187-41BD-A9AA-FE18F73BD30F} = {B924E83B-6AFE-4ABF-B79D-444F92582892}
+		{2E39F005-52FF-404D-B570-0AB2E9081272} = {B924E83B-6AFE-4ABF-B79D-444F92582892}
+		{F1F9229C-D764-4A38-BEA2-1F86A1C2FF1A} = {B924E83B-6AFE-4ABF-B79D-444F92582892}
+		{20267530-E991-4936-BFDC-0C5A1EDE90E7} = {B924E83B-6AFE-4ABF-B79D-444F92582892}
+		{CD78DA11-CB31-4892-8F96-3B7FD7015B83} = {F13F8E4E-5231-4321-9722-1BFCC08E5BB5}
+		{1874E267-5D6F-4EA4-8938-F2A0D19F27B1} = {F13F8E4E-5231-4321-9722-1BFCC08E5BB5}
+	EndGlobalSection
```
perlun, line 64: Drive-by-change, just making the Solution view in Rider/Visual Studio a bit more readable.

Perlang.Tests/ConsoleApp/ProgramTest.cs
perlun: Fixed. The problem was that the output was going to `stdout` or `stderr. I fixed support so that all output from `Program.Run()` goes to user-provided callbacks instead, making it possible to capture it from tests.

PR #90, opened 2020-10-08
(ci) Run publish job for all pushes

perlun:
...but only upload the build if the ref is `ref/head/master`.

This aims to prevent things like https://github.com/perlun/perlang/commit/ed9205a043c137fb7138fef060dcf5d3d94f3c7b,
where we merge things into master which are broken, only to detect it _after_ it has been merged.

Also fixes the publish step. The problem was that we were building an individual `.csproj`, which means that `$(SolutionDir)` will not be what you'd expect it to be.

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -8,12 +8,15 @@ jobs:
 
     steps:
       - uses: actions/checkout@v1
+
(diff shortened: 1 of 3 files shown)


perlun:
CI cannot succeed with this PR alone, but merging anyway since we see it working in #93

PR #89, opened 2020-10-07
(many) Add StyleCop linting

perlun:
Adds a basic setup of StyleCop, along with related fixes. Closes #35.﻿

Issue #35, opened 2020-03-10
Add linting via StyleCop


perlun:
This blog post seems to have some of the relevant details and not be completely outdated: https://medium.com/@michaelparkerdev/linting-c-in-2019-stylecop-sonar-resharper-and-roslyn-73e88af57ebd

PR #99, opened 2020-10-14
(many) Add SonarAnalyzer support

perlun:
Further continuing what we started with the #35-related PRs (#89 in particular), this PR adds SonarAnalyzer-based analysis as well.

I'm not overly pleased with these, to be honest, but it does one thing very well: it forbids unnecessary usings being merged by accident into master. This is the critical selling point for me that makes it worthwhile. And who knows? Maybe it'll bring in some other useful analyzers as well.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,11 +1,13 @@
 <Project>
-    <!-- StyleCop Analyzers configuration -->
+    <!-- C# analyzers configuration -->
     <PropertyGroup>
-        <CodeAnalysisRuleSet>$(SolutionDir)StyleCop.ruleset</CodeAnalysisRuleSet>
+        <CodeAnalysisRuleSet>$(SolutionDir)global.ruleset</CodeAnalysisRuleSet>
     </PropertyGroup>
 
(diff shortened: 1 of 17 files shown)


Perlang.Stdlib/Perlang.Stdlib.csproj
```diff
@@ -22,4 +22,11 @@
       <ProjectReference Include="..\Perlang.Common\Perlang.Common.csproj" />
     </ItemGroup>
 
+    <ItemGroup>
+      <PackageReference Include="SonarAnalyzer.CSharp" Version="8.14.0.22654">
+        <PrivateAssets>all</PrivateAssets>
+        <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      </PackageReference>
+    </ItemGroup>
```
perlun, line 30: Hmm, this one could & should be removed.

perlun: Fixed, resolving.

PR #106, opened 2020-10-19
(many) Make --version indicate pre-releases in a SemVer-compliant way

perlun:
Not fully happy with the weird semantics in the `ProgramTest` test method I added, but apart from that, it should be good to go. Might look one more time at that before merging it.

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -42,17 +42,17 @@ jobs:
       # Create .tar.gz archives
       #
       - name: Create linux-x64 .tar.gz file
-        run: tar cvzf ../perlang-${{steps.timestamp.outputs.timestamp}}-${GITHUB_SHA::9}-linux-x64.tar.gz *
+        run: tar cvzf ../perlang-$(./Perlang.ConsoleApp/bin/Release/netcoreapp3.1/linux-x64/perlang -v)-linux-x64.tar.gz *
         working-directory: ./Perlang.ConsoleApp/bin/Release/netcoreapp3.1/linux-x64/publish
         if: github.ref == 'refs/heads/master'
 
       - name: Create osx-x64 .tar.gz file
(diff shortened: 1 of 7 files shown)


perlun:
> Not fully happy with the weird semantics in the `ProgramTest` test method I added, but apart from that, it should be good to go. Might look one more time at that before merging it.

I ended up having to fix this to make the tests pass. #108 contains the fix.

.github/workflows/publish-packages.yml
```diff
@@ -42,17 +42,17 @@ jobs:
       # Create .tar.gz archives
       #
       - name: Create linux-x64 .tar.gz file
-        run: tar cvzf ../perlang-${{steps.timestamp.outputs.timestamp}}-${GITHUB_SHA::9}-linux-x64.tar.gz *
+        run: tar cvzf ../perlang-$(./Perlang.ConsoleApp/bin/Release/netcoreapp3.1/linux-x64/perlang -v)-linux-x64.tar.gz *
```
perlun, line 45: Example name with this change: `perlang-0.1.0-dev.95-linux-x64.tar.gz`. The Git SHA ref is not included in the filename, but it can easily be determined by running the REPL. Good enough for now, and I prefer the file name to be kept a bit shorter.

perlun: ...and, of course this didn't work because the `working-directory` is changed. :sweat_smile: I really hate it when it happens, things that get discovered _after_ the merge. A couple of things:

- Let's try to enable `set -o pipefail` here, it even went as far as publishing files with names like `/builds/perlang--linux-x64.tar.gz` to the server. Not cool in any way.
- We can skip the `if` condition here. It's fine if it creates the `.tar.gz` files here as well, we could even `ls` the files before we upload them and just skip the final step. That minimizes the "untested" parts of the CI pipeline, i.e. makes the lion share of it be tested on each MR which is always a Good Thing.<sup>&trade;</sup>

Issue #66, opened 2020-09-02
Add class support


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/66. The GitHub page you are currently reading will not contain the latest information on this issue.

---

This is not going to be trivial, partially since our current `TypeReference` implementation presumes that variables and functions always return a valid CLR type. Well, a _class_ being defined in the global (or local) scope, isn't a CLR type (at least not until we make the move to make all Perlang classes exist as valid CLR types as well - this move will come at _some_ point to make Perlang code interop with existing .NET code properly, being able to implement .NET interfaces by Perlang classes etc), so that's at least one of the challenges. There will probably be numerous others as well.

Here's the Crafting Interpreters chapter about the `jlox` implementation of classes: https://craftinginterpreters.com/classes.html. It doesn't help us _that_ much since the static typing support we added in #54 has made things much more complex. We are already quite different from `jlox` and there is unfortunately no-one apart from ourselves that we can ask for help about how things work in our code base...

## Initial features (supported in 0.1.0)
- [x] Calling `static` C# methods to support things like `Base64.encode` and `Base64.decode` (initially exposed as `base64_encode` and `base64_decode` but moved to class-based static methods in #79)

## Features likely to get implemented, but not as of the initial effort (0.2.0 or "Later")

- [ ] User-defined classes: define classes in Perlang code and be able to instantiate them.
  - This was started in the 0.1.0 milestone, but the code that existed was removed in #183. Adding it here for reference. We should put it back, and make it work equally well as in Lox.
- [ ] C#-defined classes: be able to instantiate classes defined in C#. This would be very useful and help us to add nice things like an `IntList` (mutable `List<int>`, for lack of generics)
  - We'll need a `new` keyword for this or similar, unless we go the Ruby approach and just tack `.new()` _after_ the class name. Making constructors be nothing else than plain static factory methods does have a few advantages, but it makes things like `final` fields (which must be initialized statically or in the constructor) way harder to achieve.
- [ ] Inheritance: while this is nice and a fundamental in OOP, this is a "more advanced" feature and might get implemented a bit later. We are not hostile to OOP in any way, so I am not opposing it, just deferring it slightly.
[...]

PR #113, opened 2020-11-06
(many) Move argv_pop to Argv.pop

perlun:
Now that we have parts of #66 in place, we can move more and more parts of the stdlib into proper class-based code. There are still some limitations with this though, will elaborate on this in the code diff.


Perlang.Common/IInterpreter.cs
```diff
@@ -4,8 +4,6 @@ namespace Perlang
 {
     public interface IInterpreter
     {
-        List<string> Arguments { get; }
-
         void ExecuteBlock(IEnumerable<Stmt> statements, IEnvironment blockEnvironment);
```
perlun, line 7: Given that we only have this single method left here now, we could possible get rid of this interface altogether. I might give it a try in a subsequent PR.

Perlang.Common/Perlang.Common.csproj
```diff
@@ -19,4 +19,8 @@
       <NoWarn>1591;1701;1702</NoWarn>
     </PropertyGroup>
 
+    <ItemGroup>
+      <PackageReference Include="JetBrains.Annotations" Version="2020.2.0-eap" />
```
perlun, line 23: (These make it possible to get rid of "unused class" and "unused method" warnings in IDEs like JetBrains Rider.)

Perlang.Stdlib/Stdlib/Argv.cs
```diff
@@ -0,0 +1,36 @@
...
+    [GlobalClass]
+    public static class Argv
+    {
+        private static ConcurrentQueue<string> remainingArguments;
+
+        [ArgumentsSetter]
+        public static void SetArguments(ImmutableList<string> arguments)
+        {
+            remainingArguments = new ConcurrentQueue<string>(arguments);
+        }
```
perlun, line 17: I'm not incredibly proud of this approach. Global, mutable state is [clearly evil](https://medium.com/@piotr.chmielowski/singleton-is-not-an-anti-pattern-51392c3e508e). On the other hand, we don't support invoking instance-methods at the moment from Perlang code, and even if we would, we'd still need some form of "container" object to hold the `Argv` instance.

I'm willing to say we suffer with this deficiency for now. It's the best we can do at the moment, and it's at least thread-safe (even though popping command line arguments from multiple threads concurrently would be a rather absurd programming pattern. :see_no_evil:)

perlun:
Follow-up to #79.

Issue #114, opened 2020-11-07
Support calling property getters from Perlang code


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/114. The GitHub page you are currently reading will not contain the latest information on this issue.

---

At the moment, a property getter defined like this is impossible to call from Perlang code:

```c#
    [GlobalClass]
    public static class Time
    {
        public static DateTime Now =>
            DateTime.Now;
    }
```

This method generated by this would be `Time.get_Now()`, but this method is currently impossible to call from Perlang (because of our automatic CamelCase-conversion of method invocations...). There are two problems here that I think we should solve here:

- Support calling methods with `Some_form_of_Snake_Case` from Perlang. We should check _both_ the camel-cased form of the method name as well as the "unconverted" form of the name before giving up. This would be step 1.
- Support calling static property getters. I think we go with the C# syntax here, i.e. make it possible to just call `Time.now` (or `Time.Now` if you so prefer) and be happy. No parentheses should be required, so this is what makes it slightly more complex.

PR #116, opened 2020-11-07
(stdlib) Move clock() to Time.now

perlun:
Once this is merged, we should be able to potentially get rid of `PerlangFunction` altogether, _removing_ support for global functions. :tada: (This means that we are moving more towards become an OOP language than an FP language.)

Follow-up to #79.


Perlang.Tests.Integration/Stdlib/TimeTests.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            Assert.IsType<DateTime>(Eval("Time.now()"));
+        }
+
+        // TODO: Simplify this to Time.now.ticks once we have property getter support in place
+        // TODO: https://github.com/perlang-org/perlang/issues/114
+        [Fact]
+        public void Time_now_get_Ticks_returns_a_value_greater_than_zero()
+        {
+            Assert.True((long) Eval("Time.now().get_Ticks()") > 0);
```
perlun, line 27: A quick hack here would be to make `Time.now().ticks()` work. This would probably be quite simple to achieve and we could remove that hack once we have proper property getter support in place (#114). Might give it a look.

PR #119, opened 2020-11-08
(interpreter)(common) Fix NullReferenceException

### LINKED ISSUE #117: NullReferenceException when trying to call non-existent method in .NET class
Discovered right now, this doesn't look so great:

```
> Time.now().ticks()
[line 1] Error at 'ticks': Failed to locate method 'ticks' in class 'DateTime
Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Linq.ImmutableArrayExtensions.Any[T](ImmutableArray`1 immutableArray)
   at Perlang.Interpreter.Typing.TypeValidator.TypeResolver.VisitCallExpr(Call expr)
   at Perlang.Expr.Call.Accept[TR](IVisitor`1 visitor)
   at Perlang.Interpreter.Typing.TypeValidator.TypeResolver.Validate(Expr expr)
   at Perlang.Interpreter.Typing.TypeValidator.Validate(Expr expr, Action`1 typeValidationErrorCallback, Func`2 getVariableOrFunctionCallback)
   at Perlang.Interpreter.PerlangInterpreter.Eval(String source, ScanErrorHandler scanErrorHandler, ParseErrorHandler parseErrorHandler, ResolveErrorHandler resolveErrorHandler, TypeValidationErrorHandler typeValidationErrorHandler)
   at Perlang.ConsoleApp.Program.Run(String source)
   at Perlang.ConsoleApp.Program.RunPrompt()
```

perlun:
Fixes #117

### DIFF
--- a/Perlang.Tests.Integration/Stdlib/ArgvTests.cs
+++ b/Perlang.Tests.Integration/Stdlib/ArgvTests.cs
@@ -1,13 +1,23 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Perlang.Interpreter;
+using Perlang.Stdlib;
 using Xunit;
 using static Perlang.Tests.Integration.EvalHelper;
 
 namespace Perlang.Tests.Integration.Stdlib
 {
-    public class ArgvTests
+    public class ArgvTests : IDisposable
     {
+        public ArgvTests()
+        {
+            // Motivation: The Argv [ArgumentsSetter] approach is not thread safe, and xUnit runs unit tests in parallel
+            // by default.
+            Monitor.Enter(typeof(Argv));
+        }
+
         [Fact]
         public void Argv_pop_is_defined()
         {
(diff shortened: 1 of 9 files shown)


Perlang.Tests.Integration/Stdlib/ArgvTests.cs
```diff
@@ -1,13 +1,23 @@
...
 namespace Perlang.Tests.Integration.Stdlib
 {
-    public class ArgvTests
+    public class ArgvTests : IDisposable
     {
+        public ArgvTests()
+        {
+            // Motivation: The Argv [ArgumentsSetter] approach is not thread safe, and xUnit runs unit tests in parallel
+            // by default.
+            Monitor.Enter(typeof(Argv));
```
perlun, line 18: The fact that we need to do essentially a `lock(typeof(Argv))` around our test methods which uses `Argv.pop` is really an incredibly ugly code smell. We should try to find ways around that.

Perhaps make `Argv` in "Perlang scope" be not the `Argv` class, but instead an instance of `Argv`, so that `Argv.pop` would really be calling an instance method that applies to this instance of the `PerlangInterpreter`? This would be a possibility, but this has the disadvantage of making things be more "magic" and less down-to-earth. Has to be considered carefully before we decide either way.

PR #122, opened 2020-11-09
(many) Change Argv from being static class to ARGV superglobal instance variable

perlun:
### The problem

The previous approach would introduce something *very* ugly. Global, mutable state. As we all know, this is one of the greater evils in software development. It can introduce nasty bugs, and we already saw that with the previous implementation (https://github.com/perlang-org/perlang/pull/119#discussion_r519475810).

Avoiding the global, mutable state is the proper fix to this.

### The solution

This PR introduces something I fancifully named "superglobals". It's not really something super-strange, it's just "global variables which are always available". Pretty much like [the PHP concept with the same name](https://www.php.net/manual/en/language.variables.superglobals.php). We use this to define a global variable named `ARGV` which will always contain the parameters passed to the current program.

The object is an `Argv` instance, scoped to the particular `PerlangInterpreter` instance. This is a much, much better implementation than what we previously had, where the `Argv` state was global, shared among all interpreter instances.

At the moment, the superglobal concept is not something with fancy decorators (i.e. a `[SuperGlobalAttribute]` or anything like that). It's completely tailor-made. If/when we need to add more of these, we can expand it to be more cleanly handled, but I'm not really sure at this stage what the use case for that would be.


Perlang.Common/Attributes/ArgumentsSetterAttribute.cs
```diff
@@ -1,22 +0,0 @@
...
-    /// </code>
-    ///
-    /// Note that the method might be called multiple times, if multiple Perlang programs are being started during the
-    /// lifetime of the process.
-    /// </summary>
-    [MeansImplicitUse]
-    public class ArgumentsSetterAttribute : Attribute
-    {
-    }
-}
```
perlun, line 22: This concept sure didn't last long. :see_no_evil: It was introduced in #113 and gets removed now. However, speaking about the "If you are not part of the solution, you must be part of the problem" thinking - this class was clearly part of the problem. It must be destroyed by :fire:.

perlun:
Grr, GitHub closed my PR by mistake. :grimacing: Superseded by #129.

perlun:
This definitely did **not** make it for 0.1.0, but I'll add it to that milestone anyway, to keep the list of "PRs not assigned to any milestone" empty and clean.

PR #123, opened 2020-11-09
(common)(interpreter) Fix NullReferenceException in RuntimeError handling

perlun:
Discovered while working on #122, extracted to a separate PR to keep the history clean and manageable.

### DIFF
--- a/Perlang.Common/Expr.cs
+++ b/Perlang.Common/Expr.cs
@@ -83,7 +83,7 @@ namespace Perlang
             }
         }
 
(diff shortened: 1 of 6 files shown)


perlun:
CI fails on this because of intermittent problems with the fix in https://github.com/perlang-org/perlang/pull/119#discussion_r519475810. For whatever reason, that fix is not water-proof.

Since that ugly, "here-be-dragons"-style code is going away in #122 shortly, I'll get this merged albeit the CI failure now.

Issue #125, opened 2020-11-09
Introduce immutable bindings


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/125. The GitHub page you are currently reading will not contain the latest information on this issue.

---

I realized while working on #122 that it is currently possible to mutate the super-global defined there (e.g. `ARGV = "foo"`). This is clearly not optimal.

Making it even worse, this is the same with other language elements already present in the language as well:

```
Perlang Interactive REPL Console (0.1.0-dev.108, built from 1787ac5)
> fun hello(): void { print 1; }
> hello();
1
> hello = "gurka"
gurka
> print hello; 
gurka
```

### Suggested approach

- Add a `bool IsMutable` property in the `Binding` base class (defaults to false).
- Use the property above to check if an assign expression is valid or not. Given that immutability will hopefully be an important concept in Perlang (if we do things right), I think it makes sense to to do a proper `ImmutabilityValidator` class for this right from the start.
- Expand on this as we go along, implementing more class support (#66), immutable class fields and properties, etc etc. We should also be able to utilize this in #122 pretty much right away.

perlun:
Just double-checked; the example given in https://github.com/perlang-org/perlang/issues/125#issue-739021241 is still valid. Function bindings are mutable at the moment.

perlun:
> Just double-checked; the example given in [#125 (comment)](https://github.com/perlang-org/perlang/issues/125#issue-739021241) is still valid. Function bindings are mutable at the moment.

Interestingly enough, this is no longer the case. With commit 017f2f2, it behaves like this:

```
Perlang Interactive REPL Console (0.3.0-dev.20, built from git commit 017f2f2)
Type /help for more information or /quit to quit the program.

> fun hello(): void { print 1; }
> hello();
1
> hello = "gurka"
[line 1] Error at 'hello': Function 'hello' is immutable and cannot be modified.
```

perlun:
However, this part:

> Expand on this as we go along, implementing more class support (https://github.com/perlang-org/perlang/issues/66), immutable class fields and properties, etc etc. We should also be able to utilize this in https://github.com/perlang-org/perlang/pull/122 pretty much right away.

...is not yet implemented, and I think a suitable first step would be to try and articulate an idea for how immutability could be handled in the language in a new issue. Leaving this one open until that has been done, as a reminder for myself.

One thing to think about: should we do something like this? :thinking: 

[...]

PR #127, opened 2020-11-11
(interpreter) First steps towards immutability

perlun:
This implements a significant subset of #125. Once this is done and merged, I should be able to fix #122 to handle the immutability right from the start. It simply doesn't make any sense to be able to overwrite `ARGV` with some random content, that's just plain bizarre. (and not very type-safe either...)

### DIFF
--- a/Perlang.Common/Expr.cs
+++ b/Perlang.Common/Expr.cs
@@ -42,20 +42,42 @@ namespace Perlang
             }
         }
 
+        /// <summary>
+        /// An assignment expression.
+        /// </summary>
         public class Assign : Expr
         {
-            public Token Name { get; }
+            /// <summary>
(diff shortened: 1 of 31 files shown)


Perlang.Common/Expr.cs
```diff
@@ -42,21 +42,43 @@ public override TR Accept<TR>(IVisitor<TR> visitor)
...
 
             public override TR Accept<TR>(IVisitor<TR> visitor)
             {
                 return visitor.VisitAssignExpr(this);
             }
+
+            public override string ToString()
+            {
+                return $"{Name.Lexeme} = {Value}";
+            }
```
perlun, line 81: We should add more of these. It makes it much nicer to follow along what's going on while debugging.

Issue #70, opened 2020-09-05
Support all basic integer and floating-point primitive types


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/70. The GitHub page you are currently reading will not contain the latest information on this issue.

---

I just noted that support for long (64-bit) integers is missing. This should be nothing more than adding two lines of code in the `Perlang.Interpreter.Typing.TypeValidator.TypeResolver.ResolveExplicitTypes` method, since I think the other supporting code is already there, but let's go the extra mile of adding at least one test for this as well, so we avoid trivial breakage in the future.

```perlang
> var a: long = 4 * 1024 * 1024 * 1024;
[line 1] Error at 'long': Type not found: long
```
----

Overall list of C# basic data types which are missing:

- `byte`
- `sbyte`
- `short`
- `ushort`
- `long` (#210)
- `ulong` (#337)
- `uint` (#307)
- `float` (#353)
- `double` (#236)
- `decimal`
  - Note: as #406 is being implemented, this might actually become harder since this is not supported in standard C or C++. However, `__float128` exists as a GNU extension and seems to [be supported by clang](https://discourse.llvm.org/t/error-float128-is-not-supported-on-this-target/72397), at least on some platforms. Definitely take this **after** all other types are done and consider moving it to a separate issue since 128-bit floating point is quite an edge case anyway, might be able to defer this for a few more years realistically. And if it doesn't work on e.g. FreeBSD (which seems to be the case), we might have to conditionalize this to only exist on certain platforms... which fragments the language in an extremely unpleasant way.

We may or may not want to use the exact names in Perlang, but we should probably add all or most of these at the same time.

perlun:
We should give this some focus; it's quite absurd that the current expression `var l = 2147483647; print(l.get_type());` will create an implicitly typed `uint` variable which is _impossible to specify with an explicit type_. I.e. you **must** use type inference for this; the `uint` type simply does not exist in the language yet. Even though variables can be inferred to this type... Incredibly weird semantics and we must get this fixed for 0.2.0, to avoid being ashamed of ourselves. :slightly_smiling_face:

perlun:
> We should give this some focus; it's quite absurd that the current expression `var l = 2147483648; print(l.get_type());` will create an implicitly typed `uint` variable which is _impossible to specify with an explicit type_. I.e. you **must** use type inference for this; the `uint` type simply does not exist in the language yet.

[...]

Issue #62, opened 2020-08-11
Better handling of overflows of `System.Int32` in the REPL


perlun:
This user experience cannot really be called great. We should think about some more reasonable semantics here.

```
$ perlang
Perlang Interactive REPL Console (0.1.0+ea4f649)
> 15 * 1024 * 1024 * 1024
-1073741824
> 16 * 1024 * 1024 * 1024
0
```

perlun:
Come to think of it, this is not just with the REPL; it's the same with overflows in literal initializers to explicitly typed `var` statements as well:

```perlang
> var a: int = 1 * 1024 * 1024 * 1024; print a;
1073741824
> var a: int = 2 * 1024 * 1024 * 1024; print a;
-2147483648
> var a: int = 3 * 1024 * 1024 * 1024; print a;
-1073741824
> var a: int = 4 * 1024 * 1024 * 1024; print a;
0
```

I don't think this is really good. C# gives you "Overflow in constant value computation" for example 2-4 above. That's a much better and nicer experience for the end user.

perlun:
For reference, the semantics described above is the same with Java (version 11 tested below). Given this, it cannot be that terrible. :sunglasses:  

```
$ jshell 
|  Welcome to JShell -- Version 11.0.11
|  For an introduction type: /help intro

jshell> 15 * 1024 * 1024 * 1024
$1 ==> -1073741824

jshell> 16 * 1024 * 1024 * 1024
$2 ==> 0
```

perlun:
I think this is actually fine for now. The C#/Java semantics are reasonable in this case (= allow overflow by default), since the alternative would have a extremely significant performance impact.

We might want to add some `checked` keyword at some point, but that deserves an issue of its own in that case.

PR #133, opened 2020-11-17
(many) Implement ** exponential operator

perlun:
This is mostly just "for fun". Ruby has it, and it's remotely useful in `irb` to be able to do stuff like `2 ** 32`, so why not.

One interesting detail is that this forced me to introduce some support for `BigInteger`s in the language, since `BigInteger.Pow()` is the best integer-oriented method in the .NET BCL available for these calculations. I think I'm fine with that. I'm even thinking about taking it one step further: how about automatic upscaling of integer values when (literal) evaluations don't fit in their default type? At the moment, #62 illustrates our current (evil and ugly) semantics in this area.

`16 * 1024 * 1024 * 1024` could quite easily be upscaled from an `int` to a `long` (once we have #70 in place, or technically we don't even need it for that...), and `1024 * 1024 * 1024 * 1024` could again be quite easily upscaled to a `BigInteger`. As long as we are talking about literal expression, it's not that difficult. At the very least, we should detect literal overflows, like our friends in the C# development team does it.

Non-literals are of course an altogether different story. I think we would be quite safe to not support any really fancy things in that area, at least not _yet_. :smile:

### DIFF
--- a/Perlang.Interpreter/PerlangInterpreter.cs
+++ b/Perlang.Interpreter/PerlangInterpreter.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Numerics;
 using System.Reflection;
 using Perlang.Attributes;
 using Perlang.Exceptions;
@@ -616,7 +617,7 @@ namespace Perlang.Interpreter
                 return;
             }
 
-            throw new RuntimeError(@operator, "Operands must be numbers.");
+            throw new RuntimeError(@operator, $"Operands must be numbers, not {left?.GetType().Name} and {right?.GetType().Name}");
         }
 
         private static bool IsValidNumberType(object? value)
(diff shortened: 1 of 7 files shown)


Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -906,14 +908,32 @@ public VoidObject VisitWhileStmt(Stmt.While stmt)
...
+                    }
+
                 case BANG_EQUAL:
                     return !IsEqual(left, right);
                 case EQUAL_EQUAL:
                     return IsEqual(left, right);
-            }
 
-            // Unreachable.
-            return null;
```
perlun, line 916: This was a really bad idea. You should _never_, ever write code like that, presuming that some "unreachable" code will never be reached. Unless the compiler tells you that is the case, you are more than likely wrong. (In my case, adding a new `enum` value broke this assumption.)

PR #134, opened 2020-11-18
(docs) Upgrade docfx to 2.57.2

### LINKED ISSUE #110: Upgrade docfx to 2.56.4
Now that https://github.com/dotnet/docfx/issues/5785 has been resolved, we could give 2.56.4 a try. It should be a simple update, hopefully just a matter of updating the URL in the `Makefile`. We should give it a look someday.

perlun:
Closes #110.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ docs-serve:
 	live-server _site
 
 docfx/docfx.exe:
-	wget -qO- https://github.com/dotnet/docfx/releases/download/v2.51/docfx.zip | busybox unzip - -d docfx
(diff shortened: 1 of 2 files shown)


perlun:
Seems to build cleanly, but I think I'll kick the tires a bit on this one locally first before merging.

perlun:
(and, 2.56.6 has been released so we could go with it instead)

perlun:
Let's not be afraid of change any more. Nay, let's _embrace_ it with our whole hearts. I tested this briefly, no obvious issues. Let's just merge it. If things goes horribly wrong, we can just revert it, it's not the end of the world.

PR #126, opened 2020-11-11
(ci) Build with .NET 5.0

perlun:
.NET 5 was [released yesterday](https://devblogs.microsoft.com/dotnet/announcing-net-5-0/). This PR moves CI building to the new version, and hence published versions will also bundle the .NET 5 runtime when this gets merged.

`make install` (& likely publishing in CI) currently fails on the following error, will need to look into this further some other day before this is mergable.

```
/usr/share/dotnet/sdk/5.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Publish.targets(397,5): error : Error compiling obj/Release/net5.0/linux-x64/linked/System.Diagnostics.Process.dll: Could not load type 'System.Threading.Tasks.TaskCompletionSource`1' from assembly 'System.Collections, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. [/home/per/git/perlang/Perlang.ConsoleApp/Perlang.ConsoleApp.csproj]
/usr/share/dotnet/sdk/5.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Publish.targets(397,5): error : Error: compilation failed for "obj/Release/net5.0/linux-x64/linked/System.Diagnostics.Process.dll" (0x80131522) [/home/per/git/perlang/Perlang.ConsoleApp/Perlang.ConsoleApp.csproj]
/usr/share/dotnet/sdk/5.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Publish.targets(295,5): error NETSDK1096: Optimizing assemblies for performance failed. You can either exclude the failing assemblies from being optimized, or set the PublishReadyToRun property to false. [/home/per/git/perlang/Perlang.ConsoleApp/Perlang.ConsoleApp.csproj]
```

PR #136, opened 2020-11-26
(ci) Add osx-arm64 builds to the list of supported architectures

perlun:
~Also adds (completely untested) support for `arm64` in `perlang-install`. If we are lucky, this should be all that's needed to run on Apple M1 hardware.~ This was no longer needed, because of #243.

Also, the changes made in #244 made this a lot simpler.

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -27,6 +27,7 @@ jobs:
           - linux-arm
           - linux-arm64
           - linux-x64
+          - osx-arm64
           - osx-x64
           - win-x64
 


perlun:
Hmm, this turned out to be much harder than I'd planned:

- .NET Core 3.1 [does not support `osx-arm64`](https://github.com/dotnet/core/blob/master/release-notes/3.1/3.1-supported-os.md)
- .NET 5 (which we don't support yet, #126) still [does not support `osx-arm64`](https://github.com/dotnet/core/blob/master/release-notes/5.0/5.0-supported-os.md). :grimacing:

So, I think we need to wait on the upstream support for this unfortunately. Will convert the PR to a draft to make this clearer.

## Upstream tickets tracking this

- https://github.com/dotnet/runtime/issues/43313 (Epic: Support Apple Silicon)
- https://github.com/dotnet/runtime/issues/41133 (Investigate building osx-arm64 on osx-x64)

perlun:
> When .NET 6 is here and we are ready for it, we can finally have real `osx-arm64` builds of Perlang shipped. rocket

These prerequisites should now be met. Let's see if this works now...

perlun:
Seems to work fine. Merging away. :tada:

PR #143, opened 2021-02-12
(interpreter) #124 Bug fix: calling methods in expr mode was broken in REPL

### LINKED ISSUE #124: Cannot call user-defined method without trailing semicolon in REPL
I just discovered today that this doesn't work as expected:

```
Perlang Interactive REPL Console (0.1.0-dev.103, built from ff7111d)
> fun hello(): void { print 1; }
> hello()
[line 1] Error at 'hello': Attempting to call undefined function 'hello'
> hello();
1
```

Interestingly enough though, these _do_ work. But remember that `print` is not a built-in function but a reserved word in the language for now:

```
Perlang Interactive REPL Console (0.1.0-dev.103, built from ff7111d)
> print 1
1
> print "hej"
hej
```

Calling .NET-native functions like `Base64.encode` works. I think the problem here is specifically referring to user-defined functions.

perlun:
This turned out to be quite easy in the end; for a change, I realized fairly quickly that the problem was that we special-cased expressions in an unhealthy way in `Perlang.Interpreter.PerlangInterpreter.Eval()`. Will write more in the diff view in a minute.

Closes #124.

### DIFF
--- a/Perlang.Interpreter/Immutability/ImmutabilityValidator.cs
+++ b/Perlang.Interpreter/Immutability/ImmutabilityValidator.cs
@@ -22,14 +22,6 @@ namespace Perlang.Interpreter.Immutability
             new ImmutabilityValidator(immutabilityValidationErrorCallback, getVariableOrFunctionBinding).Visit(statements);
         }
 
-        public static void Validate(
-            Expr expr,
-            Action<ValidationError> immutabilityValidationErrorCallback,
-            Func<Expr, Binding?> getVariableOrFunctionBinding)
-        {
-            new ImmutabilityValidator(immutabilityValidationErrorCallback, getVariableOrFunctionBinding).Visit(expr);
-        }
-
         private ImmutabilityValidator(
             Action<ValidationError> immutabilityValidationErrorCallback,
             Func<Expr, Binding?> getVariableOrFunctionBinding)
--- a/Perlang.Interpreter/PerlangInterpreter.cs
+++ b/Perlang.Interpreter/PerlangInterpreter.cs
@@ -301,6 +301,13 @@ namespace Perlang.Interpreter
             }
             else if (syntax is Expr expr)
             {
+                // Even though this is an expression, we need to make it a statement here so we can run the various
+                // validation steps on the complete program now (all the statements executed up to now + the expression
+                // we just received).
+                var previousAndNewStatements = previousStatements
+                    .Concat(ImmutableList.Create(new Stmt.ExpressionStmt(expr)))
+                    .ToImmutableList();
+
                 //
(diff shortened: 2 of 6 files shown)


Perlang.Interpreter/Immutability/ImmutabilityValidator.cs
```diff
@@ -22,14 +22,6 @@ internal class ImmutabilityValidator : VisitorBase
...
         }
 
-        public static void Validate(
-            Expr expr,
-            Action<ValidationError> immutabilityValidationErrorCallback,
-            Func<Expr, Binding?> getVariableOrFunctionBinding)
-        {
-            new ImmutabilityValidator(immutabilityValidationErrorCallback, getVariableOrFunctionBinding).Visit(expr);
-        }
-
```
perlun, line 32: :tada: No need for this any more now since we unified all immutability validation to use the overload which takes a `List<Stmt>` parameter.

Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -365,6 +372,12 @@ private void RegisterGlobalClasses()
                     GetVariableOrFunctionBinding
                 );
 
+                // All validation was successful, but unlike for statements, there is no need to mutate the
+                // previousStatements field in this case. Think about it for a moment. We know that the line being
+                // interpreted is an expression, so it _cannot_ have declared any new variable or anything like that
+                // (those are only allowed in statements). Hence, we presume that this expression is, if you will,
+                // "side-effect-free" in that sense.
+
```
perlun, line 380: The hunks for this file is the critical part of this PR. We simply _must_ use the information in the previous statement, together with the expression given in the new statement to be able to properly perform all validations. :point_up: is where the bug was that caused this problem.

Perlang.Interpreter/Typing/TypeValidator.cs
```diff
@@ -80,66 +80,6 @@ internal static class TypeValidator
...
-
-            // The whole expression tree should be walked by now; any type references still not resolved at this point
-            // is a critical error that should fail the type validation. To provide as much information to the user
-            // as possible, the full list errors (if any) are reported back to the caller; we don't just stop at the
-            // first error encountered. (The compiler could potentially discard information except for the first n
-            // errors if desired, though. The key point here is to not discard it at the wrong stage in the pipeline.)
-            new TypeValidatorHelper(getVariableOrFunctionCallback, typeValidationErrorCallback)
-                .ReportErrors(expr);
-        }
-
```
perlun, line 142: :rocket: Again, nice to be able to remove some code now that "everything is a statement". :joy:

Issue #140, opened 2021-02-10
Add documentation for "learning Perlang"


perlun:
## The Learn section

Right now, the docs at https://perlang.org are all fine and nice, but they lack one obvious thing: there is absolutely no _tutorials_ for "getting started with Perlang". Something like this would be useful:

- How to write an "Hello World" program with Perlang.
- How to calculate a fibonacci sequence.
- How to define a custom function and call it.
- How to use the REPL (and exit it).

Basically, describing the 0.1.0 functionality as it stands.

The idea is to use [Asciinema](https://asciinema.org/) for recording this.

## Thinking a bit more about it

Maybe the Learn section should be revised a bit. What we have there at the moment is more something like a "quick start", "for the impatient" or whatever. We should consider adding a proper "Learn" page _where we go through everything supported in the language_ as it currently stands. I think this will be good, for even I myself tend to forget what the syntax is when I haven't been working on the code lately. If that's the case for _me_, how on Earth could I expect anyone else to remember it??

perlun:
Some tutorials are in place, but I think we should aim a bit higher before considering this close. Ideally, all (or at least a significant amount) of the language elements should be covered by the tutorials.

perlun:
Maybe this is not _required_ (although it would be nice!) for the 0.1.0 release. Bumping forward to the next milestone.

PR #148, opened 2021-02-17
(many) Implement addition and subtraction assignment

perlun:
I started thinking about this when working on some tutorials for #140. It felt lame to not be able to use `+=` operators when converting [this Pi calculation](http://ajennings.net/blog/a-million-digits-of-pi-in-9-lines-of-javascript.html) from JavaScript:

```javascript
let i = 1n;
let x = 3n * (10n ** 1020n);
let pi = x;
while (x > 0) {
        x = x * i / ((i + 1n) * 4n);
        pi += x / (i + 2n);
        i += 2n;
}
console.log(pi / (10n ** 20n));
```

Adding this was fairly straightforward. I used the approach of not introducing specific AST node types for this, but just building on what's already there. There was one slight gotcha which the unit tests helped me uncover; will write some more details about it in the diff.

### Limitations

Note that *only* `+=` and `-=` is added in this PR. C has a bunch of other nice ones, like `*=`, `/=` and perhaps even some slightly more esoteric ones like `%=`. I'm not sure how far we would go (and we don't support `%` for modulo right now anyway), but _if_ we want, at least `*=` and `/=` would be trivial to implement on top of this.

### DIFF
--- a/Perlang.Interpreter/Typing/TypeValidator.cs
+++ b/Perlang.Interpreter/Typing/TypeValidator.cs
@@ -38,6 +38,9 @@ namespace Perlang.Interpreter.Typing
         {
             bool typeResolvingFailed = false;
 
+            //
+            // Phase 1: Resolve explicit and explicit type references to their corresponding CLR types.
+            //
             var typeResolver = new TypeResolver(
                 getVariableOrFunctionCallback,
                 validationError =>
@@ -49,7 +52,6 @@ namespace Perlang.Interpreter.Typing
 
             try
             {
-                // Walk the tree, resolving explicit and inferred type references to their corresponding CLR types.
                 typeResolver.Resolve(statements);
             }
             catch (TypeValidationError e)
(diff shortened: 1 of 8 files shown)


Perlang.Interpreter/Typing/TypeValidator.cs
```diff
@@ -110,15 +116,26 @@ public void Resolve(IList<Stmt> statements)
...
+                // as += is an expression and not a statement, we need to have predictable semantics for cases like
+                // this.
+                if (!expr.TypeReference.IsResolved && expr.Value.TypeReference.IsResolved)
+                {
+                    expr.TypeReference.ClrType = expr.Value.TypeReference.ClrType;
+                }
+
+                return VoidObject.Void;
+            }
+
```
perlun, line 138: This is the gotcha I mentioned in the PR description. It took me a bit of time to track this one down. The code would _work_ for many circumstances, but not in cases like `var j = i += 2`. The problem was that the type inference in the `i + 2` (implicit binary expression) was never properly propagated to the `i = i + 2` parent expression - the `Expr.Assign` instance referred to here.

The exact problem we saw here (a binary expression inside an assignment expression) was not necessarily present before the `+=` and `-=` implementation introduced in this PR; such expressions would perhaps never occur before these changes. But still, constructs like this would fail with commits like 3263a42 and 70e5ea5:

```perlang
var i = 100;
var k = (i = i + 1);
```

The above gives me an `Internal compiler error: var k initializer Perlang.Expr+Grouping inference has not been attempted` exception. With this change, the above exception is gone. :rocket: 

(The integration tests introduced by this MR highlighted the problem, so I think we are fine coverage-wise for now.)

PR #156, opened 2021-02-22
(docs) Add CI testing of examples

perlun:
I received a message from one of our users recently, making me aware that one of the examples in the newly published Learn section of our website (https://perlang.org/learn) had a syntax error.

This is not good, such thoughts ought never to happen. This PR does two things:

- Extracts the example in question (`hello_world.per`) to a separate Perlang script, making it easier to test from the commandline
- Added CI runs to ensure that this examples, and all other examples going forward will be tested in CI before a PR gets merged. This should help us avoid problems like this in the future.
----

The syntax used in the example mentioned currently behaves incorrectly (the example works, even though it should generate a syntax error). I filed a separate issue about that: https://github.com/perlang-org/perlang/issues/157

### DIFF
--- a/.github/workflows/publish-website.yml
+++ b/.github/workflows/publish-website.yml
@@ -12,7 +12,7 @@ on:
       - Perlang**
 
 jobs:
-  publish:
+  publish-website:
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -4,16 +4,18 @@ name: Website
 # for that branch.
 on:
   push:
-    branches:
-      - "*"
-      - "!master"
+    branches-ignore:
+      - master
 
 jobs:
-  rebuild:
+  build-website:
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v1
 
-      - name: Rebuild website
-        run: make docs
+      - name: Build Perlang & rebuild website
+        run: make -j all docs
+
+      - name: Ensure examples execute without errors
+        run: make docs-test-examples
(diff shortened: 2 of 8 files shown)


.github/workflows/publish-website.yml
```diff
@@ -12,7 +12,7 @@ on:
       - Perlang**
 
 jobs:
-  publish:
+  publish-website:
```
perlun, line 15: (Drive-by-renames of the jobs, to make it easier to understand what jobs relate to what CI action in logs.)

.github/workflows/website.yml
```diff
@@ -4,16 +4,18 @@ name: Website
 # for that branch.
 on:
   push:
-    branches:
-      - "*"
-      - "!master"
+    branches-ignore:
+      - master
```
perlun, line 8: Bug fix: This action was only executed for branches without a slash in the name. Hence, `foo/bar` would _not_ be tested in CI.

Makefile
```diff
@@ -33,6 +35,9 @@ docs: docfx/docfx.exe
 docs-autobuild:
 	while true; do find docs Makefile src -type f | entr -d bash -c 'scripts/time_it make docs' ; done
 
+docs-test-examples:
+	for e in docs/examples/*.per ; do Perlang.ConsoleApp/bin/Debug/net5.0/perlang $$e ; done
+
```
perlun, line 40: This is the essence of this PR: a new `make` rule which ensures that all examples in the docs are actually valid Perlang programs, which can run without errors. Any exceptions thrown from these scripts will now cause the CI job which runs this `make` rule to fail, making us aware of the problem as quickly as possible.

perlun:
This turned out to be helpful in spotting an error in #210 now: https://github.com/perlang-org/perlang/pull/210/checks?check_run_id=3774320532. Would probably have been merged without noticing if we hadn't had this CI check in place. :pray:

PR #161, opened 2021-02-28
(consoleapp) Fix bug with parsing of incomplete assignments

perlun:
Certain incomplete statements were incorrectly being considered valid, as a result of https://github.com/perlang-org/perlang/pull/34 being merged (almost a year ago). The special-casing in `ParseError` needed to be removed after that was merged, but was incorrectly retained. This PR fixes that.

### DIFF
--- a/Perlang.ConsoleApp/Program.cs
+++ b/Perlang.ConsoleApp/Program.cs
@@ -334,14 +334,6 @@ namespace Perlang.ConsoleApp
 
         private void ParseError(ParseError parseError)
         {
-            if (parseError.ParseErrorType == ParseErrorType.MISSING_TRAILING_SEMICOLON)
-            {
-                // These errors are ignored; we will get them all them when we try to parse expressions as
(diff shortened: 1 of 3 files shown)


Perlang.ConsoleApp/Program.cs
```diff
@@ -334,14 +334,6 @@ private void Error(Token token, string message)
...
         private void ParseError(ParseError parseError)
         {
-            if (parseError.ParseErrorType == ParseErrorType.MISSING_TRAILING_SEMICOLON)
-            {
-                // These errors are ignored; we will get them all them when we try to parse expressions as
-                // statements.
-                hadError = true;
-                return;
-            }
-
```
perlun, line 344: :point_up: This is the essence of the bug fix. It would allow _other_ cases of missing trailing semicolons (like `hej hej`) to slip through. The special-casing doesn't seem to be needed any more anyway, so we'd better get rid of it.

Issue #162, opened 2021-03-09
Unhandled errors should fail-fast and halt script execution


perlun:
The following program:

```
#!/usr/bin/env perlang

print(ARGV.pop());
```

...currently prints the following output:

```
$ perlang foo.per 
[line 3] No arguments left
nil
```

This is indeed very incorrect; execution of the program should stop immediately once an unhandled exception has occurred.

perlun:
A `git bisect` session pointed me towards this commit as the probable cause of this bug:

<pre>
<a href="https://github.com/perlang-org/perlang/commit/4648b8416ee0ebee57a5805ecc17239b525e6dd4">4648b8416ee0ebee57a5805ecc17239b525e6dd4</a> is the first bad commit
commit 4648b8416ee0ebee57a5805ecc17239b525e6dd4
Author: Per Lundberg <perlun@xxx>
Date:   Fri May 15 22:56:08 2020 +0300

    (many) Store the TypeReference in the Resolver (#42)

 Perlang.Common/Expr.cs          |  9 -----
 Perlang.Common/Stmt.cs          |  8 +++--
 Perlang.Common/TypeReference.cs | 77 +++++++++++++++++++++++++++++++++++++++++
 Perlang.Interpreter/Resolver.cs | 41 ++++++++++++++--------
 Perlang.Parser/PerlangParser.cs |  4 +--
 scripts/generate_ast_classes.rb | 43 +++++------------------
 6 files changed, 120 insertions(+), 62 deletions(-)
 create mode 100644 Perlang.Common/TypeReference.cs
</pre>

I'll have to dig into the details of that commit a bit later, to try and understand why it broke this.

perlun:
Hmm... apparently, doing a `git bisect` in your lunch break doesn't seem to have been such a great idea after all. :upside_down_face: The above commit _does not_ introduce the regression after all. I'll keep on digging.

perlun:
Alright, I'm quite sure b8bca3a8d51122580ca44f610dd2bdba0f5706a5 introduced the regression. Will see if I can come up with a fix now.

PR #163, opened 2021-03-10
(interpreter) Catch and handle SystemExceptions gracefully

perlun:
This was investigated as a "bi-product" of #162. The same problem exists with operations like this, but before this PR, they would kill the whole Perlang interpreter IIRC (because of the exception thrown in the first statement).

Once we have #162 under control, we could consider adding a unit test for the error below, to ensure that execution halts correctly.

`1 / 0; print "hejhej";`

### DIFF
--- a/Perlang.Common/Expr.cs
+++ b/Perlang.Common/Expr.cs
@@ -81,7 +81,7 @@ namespace Perlang
             }
         }
 
-        public class Binary : Expr
+        public class Binary : Expr, ITokenAwareExpr
         {
             public Expr Left { get; }
             public Token Operator { get; }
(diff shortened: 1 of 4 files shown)


Perlang.Common/Expr.cs
```diff
@@ -81,7 +81,7 @@ public override string ToString()
             }
         }
 
-        public class Binary : Expr
+        public class Binary : Expr, ITokenAwareExpr
```
perlun, line 84: (We should really strive for making _all_ expressions be token-aware, so we can produce proper error messages. Non-critical for now so will ignore it for the time being.)

PR #167, opened 2021-03-14
(src) #82 Move Perlang projects to the src/ directory

### LINKED ISSUE #82: Move all Perlang projects to src/
At the moment, all the .NET projects are housed right in the root of the project:

- Perlang
- Perlang.Common
- Perlang.ConsoleApp
- Perlang.Interpreter
- Perlang.Parser
- Perlang.Stdlib
- Perlang.Tests
- Perlang.Tests.Integration

It would make more sense to just have the `.sln` file in the root of the project, but all the above under a `src/` subdirectory instead. This is how e.g. [DocFX](https://github.com/dotnet/docfx/) does it, and I think it does make sense.

perlun:
Closes #82.


perlun:
This one should be quite close, but CI is likely to fail since I haven't looked through the GitHub Actions workflow files yes.

PR #169, opened 2021-03-16
(ci) Set DocFx source branch

perlun:
I discovered when looking at https://github.com/perlang-org/perlang.org/commit/a1e93f462faf438df77e699d1571ab0ca518ea03 that the "improve this doc" links were broken. DocFx seem to have been unable to determine the branch name, in line with its rules as defined here: https://dotnet.github.io/docfx/tutorial/docfx_getting_started.html#42-set-branch-name

By setting the `DOCFX_SOURCE_BRANCH_NAME` environment variable manually, we should be able to work around this.

### DIFF
--- a/.github/workflows/publish-website.yml
+++ b/.github/workflows/publish-website.yml
@@ -18,6 +18,8 @@ jobs:
 
       - name: Rebuild website
         run: make docs
+        env:
+          DOCFX_SOURCE_BRANCH_NAME: ${{ env.GIT_BRANCH }}
 
       - name: Deploy to perlang.org repo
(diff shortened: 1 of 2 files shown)


perlun:
Hmm, doesn't seem to work correctly; https://github.com/perlang-org/perlang.org/commit/94142f3a4ea9013696e2e746383e224b0a6aa069 still includes the wrong links. :unamused:

perlun:
Managed to sort out in #170, so should be fine now. https://github.com/perlang-org/perlang.org/commit/13798f8a2caa8dd04f4e9a3370700864708c0826 looks much better as well.

Issue #174, opened 2021-03-23
Add mode switcher & dark DocFx theme


perlun:
https://github.com/steffen-wilke/darkfx/issues/1 should have the info required to implement this.

perlun:
This has been mostly accomplished as of #175 being merged yesterday, but there are still some issues to sort out before we can consider this closed:

- The `.js` improvements (to detect browser preferences) should be upstreamed to https://github.com/perlun/darkerfx properly.
- A UI glitch in the show/hide TOC can be seen on mobile when dark mode is enabled (screenshot below)
- There is currently no way to access the "theme switch" on mobile devices. This should be fixed. We can check how docs.microsoft.com has done it.

![image](https://user-images.githubusercontent.com/630613/113124450-32917600-921e-11eb-9960-90b9591ea5bb.png)

perlun:
> The `.js` improvements (to detect browser preferences) should be upstreamed to https://github.com/perlun/darkerfx properly.

Fixed in https://github.com/perlun/darkerfx/commit/5bd8d973265bd42906cdedc4afb61c1304f17e88

> A UI glitch in the show/hide TOC can be seen on mobile when dark mode is enabled (screenshot below)

This bug is btw also present in `darkfx`, so we could at the very least submit an issue to them about it at some point. Not yet fixed but I've started looking into it.

I also found another horribly looking bug, only visible on mobile:

![image](https://user-images.githubusercontent.com/630613/113282396-84560100-92ef-11eb-9654-d47b37c425d2.png)

perlun:
> I also found another horribly looking bug, only visible on mobile. If you click Show / Hide Table of Contents one or more times, you can run into this:

Not reproducible on iOS with a "real" mobile device, only with iPhone simulator in Firefox DevTools. Ignoring this bug for now as long as we cannot reproduce it on mobile.

perlun:
> > A UI glitch in the show/hide TOC can be seen on mobile when dark mode is enabled (screenshot below)

Fixed in https://github.com/perlang-org/perlang/pull/176; upstreamed to darkerfx in https://github.com/perlun/darkerfx/pull/2

> This bug is btw also present in `darkfx`, so we could at the very least submit an issue to them about it at some point. Not yet fixed but I've started looking into it.

https://github.com/steffen-wilke/darkfx/issues/9

perlun:
[...]

PR #179, opened 2021-04-09
(docs) Make <pre> code be readable in dark mode also

perlun:
This is an additional fix for #174. Without this, the text used #333 on black (#000) background, which is rather hard to read, to say the least.

This change retains the #333 color in light mode (where it makes sense), but changes to use `inherit` in dark mode, which looks much better.


.github/workflows/website.yml
```diff
@@ -14,6 +14,9 @@ jobs:
     steps:
       - uses: actions/checkout@v1
 
+      - name: Install SASS
+        run: npm install -g sass
+
```
perlun, line 19: The CI failed in https://github.com/perlang-org/perlang/pull/179/checks?check_run_id=2308969743; this change is here to try and fix that. I don't really understand _why_ it tried to rebuild the CSS stylesheet though, but I'm guessing the timestamp of the `.scss` file was never than the `.css` file in this case. It has worked until now, probably because the order of the timestamps was the reverse in all CI runs until this one.

PR #180, opened 2021-04-10
(many) Mark certain keywords as reserved for future use

### LINKED ISSUE #178: (parser) Define reserved words
We are close to the first "public" release of Perlang, namely 0.1.0. :tada: Before this happens, I would like to reserve a few keywords that we may or may not want to use in 0.2.0 or some other future versions of Perlang (see [reserved words](https://en.wikipedia.org/wiki/Reserved_word) in WP):

### Currently half-working; should fail faster

* `class` - this is already reserved, but it does slightly more than it should (you are expected to write `class SomeClass {}`; it's even possible to define methods in a class. There is no way to _instantiate_ the class or call these methods though). This is just outright confusing to our potential users and we would be better off by just throwing a parse error whenever the `class` keyword is encountered for now.

### Type names

At the moment, the only two ways to define a variable `foo` is one of the following forms:

```perlang
// First form
var foo = "bar";

// Second form
var foo: String = "bar";
```

People coming from TypeScript will have no problems with this, but our friends coming from the Java or C# communities will find it a bit annoying at first (since the `var` form is admittedly more clumsy to type _and_ a bit more verbose on the reader).

What if we had a third form as well?

```perlang
// Third form
string foo = "bar";
```

I do think this would add a bit of value and it wouldn't really pollute the language _that_ much. (I might regret writing this, a few years from now...)

Anyway, to be ready for this I think we should reserve some keywords for this already now, to avoid people using them for e.g. function/variable names. Something like this:

* `byte` (unsigned 8-bit)
* `sbyte` (signed 8-bit)
* `short`
* `int`
* `long`
* `ushort`
* `uint`
* `ulong`
* `float`
* `double`
* `decimal`
* `char`
* `string`

Unsurprisingly enough, this exactly matches the C# datatypes. They are good, so there's little reason to deviate from them in this case. One can of course argue about _"why is `byte` unsigned when al the other types are signed by default"_, but really, unsigned 8-bit values are much more commonplace than signed 8-bit numbers. The other ones were probably "copied from Java", semantics-wise. :wink: (which in turn copied it from C and C++, where integers are signed by default)

Note that **all** of these should be immutable by default.

### Visibility, static/instance
* `public`
* `private`
* `protected`
* `internal`
* `static`
* `volatile`

### Standard functions
(It's of course questionable if these really should be _keywords_ in some future Perlang version, but... let's see. They are so commonly used that it makes a least a slight bit of sense.)

* `printf`
* `sprintf`

### Flow control

* `switch`
* `break`
* `continue`

### Exception handling

* `try`
* `catch`
* `finally`

### Asynchronous programming

* `async`
* `await`

### Locking/synchronization

* `lock`
* `synchronized`

### Others

* `new` (for class instantiaton)
* `mut` (_maybe_)
* `let` (we'll probably end up using this as the "immutable variable binding" in the long run, so `int i = 1` described above is perhaps essentially more a short-hand for `let i: int = 1` than `var i: int = 1`)
* `const` (unsure if this is needed but better reserve a few extra ones at this early point)
* `struct`
* `enum`
* `sizeof`
* `nameof`
* `typeof`
* `asm`

perlun:
~I started thinking about this the other day. Shipping version 0.1.0 of Perlang with a half-baked `class` implementation makes little sense, if any, at all. It's much better to just make `class` a reserved word for now, clearly implying that it's likely to come in a future release but not giving the completely false impression that it's already here (when it truly isn't).~ The above is still valid, but the removal of `class` as a reserved word was extracted to #183.

I also added a bunch of other reserved words for now, inspired by C and Java/C#. There might be things here that are not present; this is not a problem, we'll just add them to 0.2.0. Hopefully this list will shrink as time passes (and we implement some of these into the language proper). In a distant future when we ship 1.0, I hope this list would be to close to empty, unless we know with a _very_ high degree of certainty that we have things we will want to use for something useful in a future release.

Closes #178

### DIFF
--- a/src/Perlang.Parser/PerlangParser.cs
+++ b/src/Perlang.Parser/PerlangParser.cs
@@ -255,7 +255,15 @@ namespace Perlang.Parser
 
         private Stmt VarDeclaration()
         {
+            if (Check(RESERVED_WORD))
+            {
+                // Special-case to provide a more helpful error message when e.g. 'byte' is being used as a variable name.
+                throw Error(Advance(), "Reserved word encountered");
+            }
+
             Token name = Consume(IDENTIFIER, "Expecting variable name.");
+
+            BlockReservedIdentifiers(name);
 
             // Support optional typing on this form:
             // var s: String;
@@ -311,6 +319,8 @@ namespace Perlang.Parser
             Consume(LEFT_PAREN, "Expect '(' after " + kind + " name.");
             var parameters = new List<Parameter>();
 
+            BlockReservedIdentifiers(name);
+
             if (!Check(RIGHT_PAREN))
             {
                 do
@@ -320,7 +330,17 @@ namespace Perlang.Parser
                         Error(Peek(), "Cannot have more than 255 parameters.");
                     }
 
-                    var parameterName = Consume(IDENTIFIER, "Expect parameter name.");
+                    if (Check(RESERVED_WORD))
+                    {
+                        // Special-case to provide a more helpful error message when e.g. 'byte' is being used as
+                        // parameter name.
+                        throw Error(Advance(), "Reserved word encountered");
+                    }
+
+                    Token parameterName = Consume(IDENTIFIER, "Expect parameter name.");
+
+                    BlockReservedIdentifiers(parameterName);
+
                     Token parameterTypeSpecifier = null;
 
                     // Parameters can optionally use a specific type. If the type is not provided, the compiler will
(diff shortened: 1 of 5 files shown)


src/Perlang.Parser/Scanner.cs
```diff
@@ -34,13 +33,16 @@ public class Scanner
                 { "this", THIS },
                 { "true", TRUE },
                 { "var", VAR },
-                { "while", WHILE }
+                { "while", WHILE },
+
+                // Reserved keywords
+                { "class", RESERVED_WORD }, // Pending #66 to be resolved.
```
perlun, line 39: This PR turned out to be way harder than I first thought (are we surprised that this is happening _again_? :wink:). I had initially intended to add all the reserved words here, but this causes problems since some of them (`int` and `string` in particular) are actually usable as type names for variables and return types. By making them be `RESERVED_WORD` here, the parser will fail to build a valid syntax tree for these use cases.

The solution is to add only the other reserved words here, and handle the type names specially, namely...

src/Perlang.Parser/PerlangParser.cs
```diff
@@ -757,5 +742,93 @@ private void Synchronize()
...
+                //case "ushort":
+                //case "uint":
+                //case "ulong":
+                case "float":
+                //case "double":
+                //case "decimal":
+                //case "char":
+                case "string":
+                    throw Error(token, "Reserved keyword encountered", ParseErrorType.RESERVED_WORD_ENCOUNTERED);
+            }
```
perlun, line 831: ...here (continuing from https://github.com/perlang-org/perlang/pull/180/files#r611243898).

Note a few things here though (I'm discovering this as I'm typing): only the type-related reserved words _for existing types_ should go here. Everything else should probably go in the `Scanner` list after all.

Anyway, I think it's best to go the longer route here and add new unit test methods for _each and every reserved keyword_. This has (at least) the following two benefits:

- It ensures that they all work as intended.
- It makes things more fine-grained. If/when we are ready to remove a reserved keyword, either because it's making it into the language proper or because we're scrapping the idea altogether, we'll just remove the test for that keyword in the same MR.

So, this will be cleaned up before I consider this PR ready for being merged.

perlun: Class-part of this MR extracted to #183 and merged. I'll deal with the remaining unit tests now.

PR #187, opened 2021-04-17
(interpreter) Extract TypeValidator inner classes

perlun:
This does not change any semantics; only the structure and some comments are changed in this.


src/Perlang.Interpreter/Typing/TypeResolver.cs
```diff
@@ -0,0 +1,513 @@
...
+namespace Perlang.Interpreter.Typing
+{
+    /// <summary>
+    /// Class responsible for resolving implicit and explicit type references.
+    ///
+    /// The class implements the Visitor pattern, using the mechanisms provided by the base class to reduce the
+    /// amount of boilerplate code. The tree traversal must be done depth-first, since the resolving for tree nodes
+    /// closer to the top are sometimes dependent on child nodes' having their type references already resolved.
+    /// </summary>
+    internal class TypeResolver : VisitorBase
```
perlun, line 18: Note that this _does_ change the visibility for these packages from `private` to `internal`. An interesting detail: `internal` in Java is _internal to the package_, but internal in C# means _internal to the project_. Both are arguably useful; ideally, the programmer would have a chance to choose which one they prefer.

One thing that Java definitely lacks vs C# is `InternalsVisibleTo`. This provides a convenient way to circumvent the `internal` keyword _for certain assemblies_ that you choose to expose your types to. This can be really useful for unit testing; you can allow them access to internal types and/or methods, without breaking encapsulation per se.

perlun: (I added a note about this to https://github.com/perlang-org/perlang/issues/178 so we don't forget it.)

PR #188, opened 2021-04-18
(interpreter) Detect reassignment from incoercible values

perlun:
A typical example of an incoercible assignment is this:

```perlang
var i = 1;
i = "foo";
```

...in other words, an integer value being reassigned a `string` value.

Other use cases are more tricky. For example, should we or should we not allow `nil` assignment by default for reference types? We'll try to decide on this before merging this PR.

### DIFF
--- a/src/Perlang.ConsoleApp/Program.cs
+++ b/src/Perlang.ConsoleApp/Program.cs
@@ -57,7 +57,7 @@ namespace Perlang.ConsoleApp
         private readonly PerlangInterpreter interpreter;
         private readonly Action<string> standardOutputHandler;
         private readonly Action<string> standardErrorHandler;
-        private readonly HashSet<WarningType> disabledWarningsAsErrors = new();
+        private readonly HashSet<WarningType> disabledWarningsAsErrors;
 
         private bool hadError;
         private bool hadRuntimeError;
@@ -85,6 +85,23 @@ namespace Perlang.ConsoleApp
             var detailedVersionOption = new Option("-V", "Show detailed version information");
             var evalOption = new Option<string>("-e", "Executes a single-line script") { AllowMultipleArgumentsPerToken = false, ArgumentHelpName = "script" };
             var printOption = new Option<string>("-p", "Parse a single-line script and output a human-readable version of the AST") { ArgumentHelpName = "script" };
+            var noWarnAsErrorOption = new Option<string>("-Wno-error", "Treats specified warning as a warning instead of an error.") { ArgumentHelpName = "error" };
+
+            var disabledWarningsAsErrorsList = new List<WarningType>();
+
+            noWarnAsErrorOption.AddValidator(result =>
+            {
+                string warningName = result.GetValueOrDefault<string>();
+
+                if (!WarningType.KnownWarning(warningName))
+                {
+                    return $"Unknown warning: {warningName}";
+                }
+
+                disabledWarningsAsErrorsList.Add(WarningType.Get(warningName));
+
+                return null;
+            });
 
             // Note: options must be present in this list to be valid for the RootCommand.
             var options = new[]
@@ -92,7 +109,8 @@ namespace Perlang.ConsoleApp
                 versionOption,
                 detailedVersionOption,
                 evalOption,
-                printOption
+                printOption,
+                noWarnAsErrorOption
             };
 
             var rootCommand = new RootCommand
@@ -131,7 +149,8 @@ namespace Perlang.ConsoleApp
 
                         var program = new Program(
                             replMode: true,
-                            standardOutputHandler: console.Out.WriteLine
+                            standardOutputHandler: console.Out.WriteLine,
+                            disabledWarningsAsErrors: disabledWarningsAsErrorsList
                         );
 
                         int result = program.Run(source, program.CompilerWarning);
@@ -150,7 +169,8 @@ namespace Perlang.ConsoleApp
 
                         new Program(
                             replMode: true,
-                            standardOutputHandler: console.Out.WriteLine
+                            standardOutputHandler: console.Out.WriteLine,
+                            disabledWarningsAsErrors: disabledWarningsAsErrorsList
                         ).ParseAndPrint(source);
 
                         return Task.FromResult(0);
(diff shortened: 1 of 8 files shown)


src/Perlang.Tests.Integration/Typing/TypingTests.cs
```diff
@@ -124,6 +124,72 @@ public void var_declaration_with_initializer_correctly_infers_type_from_assignme
...
+                var s = ""foo"";
+                s = nil;
+
+                print s;
+            ";
+
+            var output = EvalReturningOutputString(source);
+
+            Assert.Equal("nil", output);
+        }
```
perlun, line 174: > Other use cases are more tricky. For example, should we or should we not allow `nil` assignment by default for reference types? We'll try to decide on this before merging this PR.

In other words, this. :point_up: We need to figure out whether this is what we want or not. Having it like this is the easiest way out, but it should actually be quite simple to forbid all `nil` assignments also if we like. We could actually give it a try and see how many tests it breaks. :bulb:

It's probably much easier to forbid `nil` assignment at this early stage in the Perlang development than trying to get rid of it after-the-fact, which experience from other languages has proven to be quite hard (not to say _impossible_). (Well, C# has made it possible to default to non-nullability by default which is quite an achievement.)

perlun: The following is now in place:

- Reassignments of variables to `nil` emits a warning.
- Using `nil` as a variable initializer likewise emits a warning
- Bug fix: Using `nil` for local variable inference (`var s = nil`) is now properly detected as an invalid condition, causing a compilation error. There is simply _no way_ for the compiler to determine what type the user was intending to use in that case, so it's better to just fail fast.
- Passing `nil` as a function parameter also emits a warning.

The PR got a bit more complex because of this. I'll try to extract the infrastructure for emitting warnings into a separate PR in a day or two, since it's an isolated piece of art that doesn't necessarily need to be glued together with the rest like this.

One big caveat at the moment is also the following semantic detail:

- In REPL mode, warnings are treated as warnings and just produce some noise; the code will still run as if the warning wasn't there.
- In script mode, *warnings are treated as errors* (imagine `-Werror` being enabled with `gcc`) and _there is no way to turn it off_. The latter might be a bit unpleasant and we could try to fix a `-Wno-error`-flag or something when we have the infra for warning extracted to an isolated PR.

perlun: > One big caveat at the moment is also the following semantic detail:
>
> - In REPL mode, warnings are treated as warnings and just produce some noise; the code will still run as if the warning wasn't there.
> - In script mode, *warnings are treated as errors* (imagine `-Werror` being enabled with `gcc`) and _there is no way to turn it off_. The latter might be a bit unpleasant and we could try to fix a `-Wno-error`-flag or something when we have the infra for warning extracted to an isolated PR.

This was now fixed. I added a `-Wno-error=nil-usage` flag that you can use to demote usage of `nil` to just a warning instead of an error.

The REPL still always treats it as a warning which makes sense in my world view of how a REPL should "feel" to the user.

One caveat still: the `EvalHelper` methods will consider all warnings _warnings_ by default, not errors. We could change this (and perhaps we should) but it does berak about 5 unit tests. Some of them could easily be fixed but not others. We would probably have to figure out a way to disable/enable warnings for individual tests to accomplish that (perhaps by moving them to be `Program` integration tests and passing in the proper `-Wno-error` flag). But, that's a different story for another day.

> I'll try to extract the infrastructure for emitting warnings into a separate PR in a day or two, since it's an isolated piece of art that doesn't necessarily need to be glued together with the rest like this.

This part I'll try to deal with before we get this merged though.

perlun: > This part I'll try to deal with before we get this merged though.

For reference, infrastructure was extracted to #190 and #191, both of which are merged by now.

src/Perlang.ConsoleApp/Program.cs
```diff
@@ -85,14 +85,32 @@ public static int MainWithCustomConsole(string[] args, IConsole console)
             var detailedVersionOption = new Option("-V", "Show detailed version information");
             var evalOption = new Option<string>("-e", "Executes a single-line script") { AllowMultipleArgumentsPerToken = false, ArgumentHelpName = "script" };
             var printOption = new Option<string>("-p", "Parse a single-line script and output a human-readable version of the AST") { ArgumentHelpName = "script" };
+            var noWarnAsErrorOption = new Option<string>("-Wno-error", "Treats specified warning as a warning instead of an error.") { ArgumentHelpName = "error" };
+
+            var disabledWarningsAsErrorsList = new List<WarningType>();
```
perlun, line 90: (Silly `List` suffix to avoid name clash with non-static field. It's a shame the C# compiler uses the same scope for these, since a _static method cannot access the fields anyway_. Let's try to do better in Perlang when we implement support for `static` methods. :sunglasses:)

src/Perlang.Interpreter/Typing/TypesResolvedValidator.cs
```diff
@@ -408,5 +418,18 @@ public override VoidObject VisitVarStmt(Stmt.Var stmt)
...
             return VoidObject.Void;
         }
+
+        public override VoidObject VisitAssignExpr(Expr.Assign expr)
+        {
+            base.VisitAssignExpr(expr);
+
+            if (expr.Value.TypeReference.IsNullObject)
+            {
+                // TODO: Use expr.Value.Token here instead of expr.name, #189
```
perlun, line 428: #189

src/Perlang.Interpreter/Typing/TypesResolvedValidator.cs
```diff
@@ -379,6 +384,11 @@ public override VoidObject VisitVarStmt(Stmt.Var stmt)
                             $"Cannot assign {stmt.Initializer.TypeReference.ClrType.Name} value to {stmt.TypeReference.ClrType.Name}"
                         ));
                     }
+                    else if (stmt.Initializer.TypeReference.IsNullObject)
+                    {
+                        // TODO: Use stmt.Initializer.Token here instead of stmt.name, #189
```
perlun, line 389: #189

PR #190, opened 2021-05-09
(interpreter) Add infrastructure for compiler warnings

perlun:
Extracting of most of the infrastructure code from #188, to make that PR clearer and more focused on the actual semantic changes.

PR #191, opened 2021-05-10
(consoleapp) Improve argument parsing

perlun:
This is a nice by-product of #188. While working on that PR, I ran into some bugs in the way we did our command-line parsing. The full fix is available in #188, but to keep the history nice and tidy we'll extract the relevant parts to separate PR:s like this (this one and #190).

### DIFF
--- a/src/Perlang.ConsoleApp/Perlang.ConsoleApp.csproj
+++ b/src/Perlang.ConsoleApp/Perlang.ConsoleApp.csproj
@@ -37,7 +37,7 @@
 
     <ItemGroup>
         <PackageReference Include="ReadLine" Version="2.0.1" />
-        <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20158.1" />
+        <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.21216.1" />
     </ItemGroup>
 
 </Project>
--- a/src/Perlang.ConsoleApp/Program.cs
+++ b/src/Perlang.ConsoleApp/Program.cs
@@ -82,8 +82,8 @@ namespace Perlang.ConsoleApp
         {
             var versionOption = new Option(new[] { "--version", "-v" }, "Show version information");
             var detailedVersionOption = new Option("-V", "Show detailed version information");
-            var evalOption = new Option("-e", "Executes a single-line script");
-            var printOption = new Option("-p", "Parse a single-line script and output a human-readable version of the AST");
+            var evalOption = new Option<string>("-e", "Executes a single-line script") { AllowMultipleArgumentsPerToken = false, ArgumentHelpName = "script" };
+            var printOption = new Option<string>("-p", "Parse a single-line script and output a human-readable version of the AST") { ArgumentHelpName = "script" };
 
             // Note: options must be present in this list to be valid for the RootCommand.
             var options = new[]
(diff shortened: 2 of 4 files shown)


src/Perlang.ConsoleApp/Perlang.ConsoleApp.csproj
```diff
@@ -37,7 +37,7 @@
 
     <ItemGroup>
         <PackageReference Include="ReadLine" Version="2.0.1" />
-        <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20158.1" />
+        <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.21216.1" />
```
perlun, line 40: There's an even newer (beta) version we could pull in [mentioned here](https://github.com/dotnet/command-line-api/pull/1271#issuecomment-834739025), but I don't want to do it in this PR since I'm not 100% sure that its current tests would illustrate whether the fix solves our problem or not. Postponing until we are ready with #190 and #191 and in the process of getting #188 merged; that one should have all the supporting code for properly testing the fix.

src/Perlang.ConsoleApp/Program.cs
```diff
@@ -118,37 +118,39 @@ public static int MainWithCustomConsole(string[] args, IConsole console)
...
-                        console.Error.WriteLine("Error: the -e and -p option are mutually exclusive");
-                        return Task.FromResult(ExitCodes.INVALID_ARGUMENT);
-                    }
-
                     if (parseResult.HasOption(evalOption))
                     {
-                        IEnumerable<string> arguments = parseResult.Tokens
-                            .Where(t => t.Type == System.CommandLine.Parsing.TokenType.Argument)
-                            .Select(t => t.Value);
+                        // TODO: Workaround until we have a command-line-api package with https://github.com/dotnet/command-line-api/pull/1271 included.
```
perlun, line 123: https://github.com/dotnet/command-line-api/pull/1271#issuecomment-834739025

src/Perlang.Tests/ConsoleApp/ProgramTest.cs
```diff
@@ -241,7 +241,7 @@ public void with_version_parameter_outputs_expected_value()
             public void with_eval_parameter_outputs_expected_value()
             {
                 // Arrange & Act
-                Program.MainWithCustomConsole(new[] { "-e", "print", "10" }, testConsole);
+                Program.MainWithCustomConsole(new[] { "-e", "print 10" }, testConsole);
```
perlun, line 244: The way this test was formulated relied on some quite nasty semantics in `System.CommandLine`. While we are _still_ using a quite kludgy approach now, we should be able to get rid of it when incorporating the latest changes from upstream.

This does mean that you can no longer do cool stuff like `perlang -e 10 + 10`; you will _have_ to quote the expression as `perlang -e '10 + 10'`. However, the fact that we can use a more "standard" approach to arg parsing makes this a tradeoff I'm willing to do.

src/Perlang.ConsoleApp/Perlang.ConsoleApp.csproj
perlun: For reference, I'm suspecting this is the dependency change which introduced the bug described in #199. More details can be found here: https://github.com/perlang-org/perlang/issues/199#issuecomment-894684410

PR #192, opened 2021-05-15
(docs) Fix broken affix content

perlun:
While working on #140, I discovered that this was broken. The upstream problem here is more carefully described in
https://github.com/perlun/darkerfx/issues/3.

### DIFF
--- a/docs/templates/override/styles/override.css
+++ b/docs/templates/override/styles/override.css
@@ -8,6 +8,12 @@ body {
     font-size: 16px;
 }
 
+.affix ul ul > li > a:before {
+    font-size: 16px;
+    top: 0px;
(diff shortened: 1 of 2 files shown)


docs/templates/override/styles/override.css
```diff
@@ -8,6 +8,12 @@ body {
     font-size: 16px;
 }
 
+.affix ul ul > li > a:before {
+    font-size: 16px;
+    top: 0px;
+    left: 3px;
+}
+
```
perlun, line 16: This was a slightly unrelated change, but it turned out the affix-TOC was rendered with a `>` character which was rendered with the wrong font size. The position was also slightly tweaked to make it look better.

PR #194, opened 2021-05-22
(docs) Add highlight.js support for Perlang

### LINKED ISSUE #193: Add highlight.js support for Perlang
The examples shown at https://perlang.org/perlang/learn/ (soon to be moved to https://perlang.org/perlang/learn/quickstart/) look admittedly a bit dull at the moment:

image

I don't think I personally would be very interested in learning more about a language that looked like that. You're simply so _used_ to seeing code rendered with syntax highlighting active that it feels like "the way things should be".

We should look into adding some rudimentary hljs support for Perlang. This guide hopefully provides everything we need to get started: https://highlightjs.readthedocs.io/en/latest/language-guide.html

(It doesn't need to be perfect from day 1, but it should at least give a somewhat decent rendering of the examples we provide.)

perlun:
Closes #193


docs/templates/override/styles/docfx.vendor.js
```diff
@@ -0,0 +1,38 @@
...
+// @license-end
```
perlun, line 38: This is a straight copy of the `docfx.vendor.js` file from the default template, with `highlight.js` taken out so I could swap it out for another version (fixing an important bug). More context: https://github.com/dotnet/docfx/issues/7373

PR #206, opened 2021-09-15
(many) Extract binding-related code to separate class

perlun:
The end-goal here is to improve testability, by being able to mock the `IBindingHandler` and/or the `Binding` objects it keeps track of. Unfortunately, `Binding` instances unfortunately allow a certain level of mutability, making them exceptionally annoying to work with.


perlun:
> Unfortunately, `Binding` instances unfortunately allow a certain level of mutability, making them exceptionally annoying to work with.

Hmm, I wonder if this couldn't actually be worked around by _replacing_ the binding in the internal data structure once the `TypeReference` has been resolved. I wonder if it provide enough benefits to make it worthwhile.

Nonetheless, doing that kind of change becomes also simpler now that the whole binding-related code is contained in a single `BindingHandler` class.

PR #209, opened 2021-09-25
(tests) Add ResolverTest + refactor Eval() to support its work

perlun:
Previously, tests which wanted to inspect the AST or other internal work of the interpreter needed to manually create a bunch of data structures and then call the appropriate method (like `TypeValidator.Validate()`; see `TypeValidatorTest` for an example of how awkward this is).

While the approach described above _does_ have an advantage in that it requires a less functional interpreter/compiler to be able to run the tests, it does come with a significant disadvantage as well: it's tedious, it's manual work, and it's boring. It's very unlikely that any significant amount of tests would be added when the process is so painful.

The approach introduced in this PR makes things significantly better. We now have appropriate tooling in place (the `ScanParseAndResolveSingleStatement` method) to help in writing tests like this.

----

While full E2E/integration tests (which execute a full Perlang interpreter and inspect the result or makes assertions on certain kind
of errors) are useful for many aspects of the life of an interpreter/compiler, there are particular areas where it also makes
sense to be able to "straightjacket" your code a bit more. These tests are in other words not intented to replace full integration tests but to complement them.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -9,7 +9,7 @@
         <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.321" PrivateAssets="all" />
         <AdditionalFiles Include="$(SolutionDir)stylecop.json" Link="stylecop.json" />
 
-        <PackageReference Include="SonarAnalyzer.CSharp" Version="8.14.0.22654" />
+        <PackageReference Include="SonarAnalyzer.CSharp" Version="8.17.0.26580" />
     </ItemGroup>
 
     <!--
--- a/src/Perlang.Tests/Interpreter/Resolution/ResolverTest.cs
+++ b/src/Perlang.Tests/Interpreter/Resolution/ResolverTest.cs
@@ -1,0 +1,78 @@
+using System;
+using System.Collections.Immutable;
+using System.Linq;
+using Perlang.Interpreter;
+using Perlang.Interpreter.Resolution;
+using Perlang.Parser;
+using Xunit;
+
+namespace Perlang.Tests.Interpreter.Resolution
+{
+    public class ResolverTest
+    {
+        [Fact(Skip = "Known to be broken, will be fixed shortly on another branch")]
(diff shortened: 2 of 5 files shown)


Directory.Build.props
```diff
@@ -9,7 +9,7 @@
         <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.321" PrivateAssets="all" />
         <AdditionalFiles Include="$(SolutionDir)stylecop.json" Link="stylecop.json" />
 
-        <PackageReference Include="SonarAnalyzer.CSharp" Version="8.14.0.22654" />
+        <PackageReference Include="SonarAnalyzer.CSharp" Version="8.17.0.26580" />
```
perlun, line 12: This is an attempt to fix a false positive in Sonar Analyzer: https://github.com/perlang-org/perlang/pull/209/checks?check_run_id=3709390145

(Upstream issue is https://github.com/SonarSource/sonar-dotnet/issues/3105)

src/Perlang.Tests/Interpreter/Resolution/ResolverTest.cs
```diff
@@ -0,0 +1,78 @@
...
+            ");
+
+            // Assert. We want to ensure that the proper TypeReference is being used in the VariableBindingFactory
+            // which has been created at this point, since using the TypeReference from the initializer instead of the
+            // variable will work incorrectly with e.g. integer expansion (`var l: long = 123` - it's critical that
+            // TypeReference becomes `long` here and not `int` or `short`)
+            Assert.IsType<Stmt.Var>(singleStatement);
+            Assert.True(resolver.Globals.ContainsKey("i"));
+            Assert.Equal(((Stmt.Var)singleStatement).TypeReference, ((VariableBindingFactory)resolver.Globals["i"]).TypeReference);
+        }
```
perlun, line 28: (This tests will be fixed to no longer fail in #70, but it made sense to include it here already since it serves as an example of how the new `ScanAndParse` code is used from tests.)

PR #210, opened 2021-09-28
(many) Add 'long' datatype

perlun:
Originally planned for #70, this seemed like a fun thing to work on during an evening. A few weeks later :innocent: and here we are. It ended up being a bit more complex than planned because of a bug in the interpreter and because I made a significant effort in trying to reproduce this bug with automated tests (see #206, #207, #208 and #209 - they're all semi-related to this). The end result was pretty good and this should be ready for merge, but I might break out parts of to separate PR:s to keep history cleaner.

The other (signed) datatypes should be a walk in the park once this is complete, so I might end up fixing (some of) them soon as well.


src/Perlang.Interpreter/Resolution/ResolveError.cs
```diff
@@ -2,6 +2,9 @@
 
 namespace Perlang.Interpreter.Resolution
 {
+    // TODO: Rename to NameResolutionError, but need to resolve the conflict with existing type. Could we rename that
+    // one to NameResolutionTypeValidationError, or perhaps just moving it to the Typing namespace is the right thing to
+    // do?
```
perlun, line 7: This TODO comment must be fixed properly before merge.

perlun: This is currently being dealt with as part of fixing this, separate PR coming up.

> The end result was pretty good and this should be ready for merge, but I might break out parts of to separate PR:s to keep history cleaner.

perlun: > separate PR coming up.

Was merged as #211, resolving.

src/Perlang.Interpreter/NameResolution/NameResolver.cs
```diff
@@ -425,7 +425,7 @@ public VoidObject VisitVarStmt(Stmt.Var stmt)
                 Resolve(stmt.Initializer);
             }
 
-            Define(stmt.Name, stmt.Initializer?.TypeReference ?? new TypeReference(stmt.Name));
+            Define(stmt.Name, stmt.TypeReference ?? new TypeReference(stmt.Name));
```
perlun, line 428: This is an important part of this PR: the previous behavior was a bug; it would use the type of the _initializer_ even though that could potentially be of a different type than the actual variable being defined. (Think `var l: long = 123` and you get the picture; the variable should **not** become an `int`/`short` in this case)

src/Perlang.Interpreter/Typing/TypeResolver.cs
```diff
@@ -447,6 +463,16 @@ private static ITypeReference GreaterType(ITypeReference leftTypeReference, ITyp
...
+                case TypeCode.Object:
+                    if (type == typeof(BigInteger))
+                    {
+                        // Note: this is insanely wrong and will likely cause more problems than the ones it intends to
+                        // solve...
+                        return Convert.ToDouble(Decimal.MaxValue);
+                    }
+
+                    return null;
+
```
perlun, line 475: I think we should look a bit more into this before merging. Can we provoke any obvious errors caused by the simplistic logic above? :thinking: Perhaps in a unit test or so.

perlun: Added a comment for now, resolving.

PR #211, opened 2021-10-02
(interpreter) Resolve-related renamings

perlun:
- `Perlang.Interpreter.Resolution` -> `Perlang.Interpreter.NameResolution`
- `Perlang.Interpreter.Resolution.NameResolutionError` -> `Perlang.Interpreter.Typing.NameResolutionTypeValidatorError`
- `ResolveError` -> `NameResolutionError`
- `Perlang.Interpreter.Resolution.Resolver` -> `Perlang.Interpreter.NameResolution.NameResolver`


src/Perlang.Interpreter/NameResolution/NameResolver.cs
```diff
@@ -424,7 +425,7 @@ public VoidObject VisitVarStmt(Stmt.Var stmt)
                 Resolve(stmt.Initializer);
             }
 
-            Define(stmt.Name, stmt.Initializer?.TypeReference ?? new TypeReference(stmt.Name));
+            Define(stmt.Name, stmt.TypeReference ?? new TypeReference(stmt.Name));
```
perlun, line 428: Oops, this is a bug fix that should go in but not as early as in this PR. Will remove (and perhaps remove other tests which fail because of it) before merge.

perlun: Fixed, resolving.

Issue #69, opened 2020-09-05
Add support for literal hexedecimal, binary and octal numbers


perlun:
## Goals

It should be possible to specify literal hexadecimal, binary and octal integer and long values in Perlang.

## Non-goals

Supporting mechanisms like [incbin](https://nasm.us/doc/nasmdoc3.html) in [NASM](https://nasm.us/) is out of scope for this issue.

### Implementation details

Hexadecimal integers are simple, since there is a uniform notation for them supported by most languages: `0x42` => literal hexadecimal 42, i.e. the decimal value 66. (We can either be strict and enforce a lowercase `0x` here or be relaxed and support `0X` prefixes as well; Ruby as C# seem to support both.)

Binary numbers are not supported by all languages, but C# seems to support the `0b` (or `0B`) prefix for this. We could do the same.

Octal literals are interestingly enough _not_ supported in C#. Many other languages like Ruby, Python and good-old C supports octal notation using a single `0` prefix - `0755` is treated as an octal number, equivalent to `493` in decimal notation.

However, the "plain-zero" prefix is problematic in that it is too ambiguous; it's not obvious enough to any new developer unfamiliar with your language from beforehand that `042` is in fact, not the literal number _42_ but instead _34_. Perhaps to avoid this, the Rust language chose a different approach <sub>[1](https://github.com/rust-lang/rust/issues/6563) [2](https://github.com/rust-lang/rust/pull/10243)</sub> - it uses `0o42` for octal notation.

While this might look a bit "odd" at first (for people used to simple `0`-prefixing octal numbers), it does make quite a bit of sense in being:

- more consistent with other prefixes (`0x` and `0b`)
- less confusing to beginners
- mildly annoying to people who would have preferred plain `0755` notation. They can still write Unix file permissions in a natural way, they just have to do some very minor learning.

### Suggested syntax

- `0x42` - literal hexadecimal 42 (decimal 66)
- `0b10011001` - literal binary (decimal 153)
- `0o755` - literal octal (decimal 493)

We can either implement all of these in a single PR, or do a separate PR for each notation.

PR #219, opened 2021-10-12
(parser) Add support for binary literals

perlun:
More work towards completing #69.


src/Perlang.Parser/Scanner.cs
```diff
@@ -339,32 +351,37 @@ private void Number()
...
-                    value = BigInteger.Parse(source[(start + startOffset)..current], numberStyles);
-                }
+                    Base.DECIMAL =>
+                        BigInteger.Parse(source[(start + startOffset)..current], numberStyles),
+
+                    Base.BINARY =>
+                        Convert.ToUInt64(numberCharacters, 2),
+
+                    Base.HEXADECIMAL =>
+                        // Quoting from
```
perlun, line 370: Filed an upstream issue about this (seemingly) false positive error: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3392

I'll suppress this using the approach provided in https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1515.md for now.

Issue #225, opened 2021-11-11
Move away from using the `dynamic` keyword 


perlun:
Until now, we have used `dynamic` which is convenient for some parts of the interpreter which would otherwise be a bit of a "condition-spaghetti" (to support various numeric data types).

https://github.com/perlang-org/perlang/blob/7846073a244940ed63dced0f2280cc3b7980ad50/src/Perlang.Interpreter/PerlangInterpreter.cs#L987-L1093

_However_, using `dynamic` has (at least) the following disadvantages:

1. It is a form of known "technical debt" which will be harder to convert to IL code, once we are ready for a proper compiled mode in the Perlang interpreter/compiler.
2. It can cause issues like #224. It _did_ also cause the upgrade to .NET 6 to become more complex: https://github.com/perlang-org/perlang/pull/223/files#r747380990

For these reasons, we should try to get rid of `dynamic` as early as is conveniently possible.

PR #227, opened 2021-11-16
(interpreter) Move away from relying on 'dynamic' for unary prefix/postfix operators

perlun:
This is a first step towards resolving #225. The MR fixes the `-` unary prefix and the `--`/`++` postfix operators to manually handle all types instead.

### DIFF
--- a/src/Perlang.Interpreter/PerlangInterpreter.cs
+++ b/src/Perlang.Interpreter/PerlangInterpreter.cs
@@ -603,12 +603,60 @@ namespace Perlang.Interpreter
                     return !IsTruthy(right);
 
                 case MINUS:
-                    // Using 'dynamic' here is arguably a bit weird, but like in VisitBinaryExpr(), it simplifies things
-                    // significantly. The other option would be to handle all kind of numeric types here individually,
-                    // which is clearly doable but a bit more work. For now, the CheckNumberOperand() method is the
-                    // guarantee that the dynamic operation will succeed.
                     CheckNumberOperand(expr.Operator, right);
-                    return -(dynamic?)right;
+
+                    // This was previously done using the 'dynamic' keyword, but a major drawback of doing it like this
+                    // is that it makes it harder for the compiler to do a static analysis of the code. This particular
+                    // section was fine, but other parts broke as part of the .NET 6 upgrade. More details:
+                    // https://github.com/perlang-org/perlang/pull/223/files#r747380990
+                    //
+                    // The general conclusion was that it's safer and more predictable to move away from 'dynamic'
+                    // altogether.
+
+                    switch (right)
+                    {
+                        case SByte value:
+                            return -value;
+
+                        case Int16 value:
+                            return -value;
+
+                        case Int32 value:
+                            return -value;
+
+                        case Int64 value:
+                            return -value;
+
+                        case Byte value:
+                            return -value;
+
+                        case UInt16 value:
+                            return -value;
+
+                        case UInt32 value:
+                            return -value;
+
+                        case UInt64 value:
+                            // .NET will give an "ambiguous invocation" here, suggesting that we cast value to either a
+                            // `decimal`, `double` or `float`. However, making an integer-type of variable become a
+                            // floating point value all of a sudden seems like odd semantics here. Even though it is
+                            // less performant, we go with the BigInteger approach for now: negating a UInt64 will
+                            // inherently expand it to a BigInteger.
+                            return -new BigInteger(value);
+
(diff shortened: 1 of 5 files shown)


src/Perlang.Interpreter/Extensions/TypeExtensions.cs
```diff
@@ -9,6 +9,12 @@ namespace Perlang.Interpreter.Extensions
     internal static class TypeExtensions
     {
         public static string ToTypeKeyword(this Type type)
+        {
+            return ToTypeKeywordNew(type);
+        }
+
+        // TODO: replace usage of the above with this one
+        public static string ToTypeKeywordNew(Type type)
```
perlun, line 17: This mess is something we should clean up, preferably within the scope of this MR. We could also look into merging it with `Utils.StringifyType()` which has some semi-related logic as well.

In the end, we should have **one** stastic method for this and it should have perhaps two overloads:

- One that takes an `object obj` parameter, which can be nullable.
- One that takes a `Type type` parameter. (Will this even be needed? Time will tell.)

src/Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -603,12 +604,60 @@ private void Interpret(IEnumerable<Stmt> statements)
...
+
+                        case Double value:
+                            return -value;
+
+                        case BigInteger value:
+                            return -value;
+
+                        default:
+                            // TODO: Use TypeExtensions.ToTypeKeywordNew() instead. Just need to make it handle null objects
+                            throw new RuntimeError(expr.Operator, $"Internal runtime error: Unexpected type of object ${right?.GetType().FullName} encountered");
```
perlun, line 659: https://github.com/perlang-org/perlang/pull/227/files#r750651560

src/Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -636,23 +685,30 @@ public object VisitUnaryPostfixExpr(Expr.UnaryPostfix expr)
...
+                    MINUS_MINUS => previousLongValue - 1,
+                    _ => throw new RuntimeError(expr.Operator, $"Unsupported operator encountered: {expr.Operator.Type}")
+                },
+                double previousDoubleValue => expr.Operator.Type switch
+                {
+                    PLUS_PLUS => previousDoubleValue + 1,
+                    MINUS_MINUS => previousDoubleValue - 1,
+                    _ => throw new RuntimeError(expr.Operator, $"Unsupported operator encountered: {expr.Operator.Type}")
+                },
+                _ => throw new RuntimeError(expr.Operator, $"Internal runtime: Unsupported type {ToTypeKeywordNew(left?.GetType())} encountered with {expr.Operator.Type} operator")
```
perlun, line 710: Should also use https://github.com/perlang-org/perlang/pull/227/files#r750651560

src/Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -747,7 +803,7 @@ private static void CheckNumberOperands(Token @operator, object? left, object? r
                 return;
             }
 
-            throw new RuntimeError(@operator, $"Operands must be numbers, not {left?.GetType().Name} and {right?.GetType().Name}");
+            throw new RuntimeError(@operator, $"Operands must be numbers, not {(left?.GetType() ?? null!).ToTypeKeyword()} and {(right?.GetType() ?? null).ToTypeKeyword()}");
```
perlun, line 806: https://github.com/perlang-org/perlang/pull/227/files#r750651560

src/Perlang.Parser/Scanner.cs
```diff
@@ -356,6 +356,9 @@ private void Number()
 
             if (isFractional)
             {
+                // TODO: This is a mess. We currently treat all floating point values as _double_, which is insane. We
+                // TODO: should probably have a "use smallest possible type" logic as below for integers, for flotaing point
+                // TODO: values as well. We could also consider supporting `decimal` while we're at it.
```
perlun, line 361: See #70. I think that adding an explicit `float` and `double` should get fairly high priority, since the current way (of just putting everything in there as `double`s) is extremely messed up, to be honest.

src/Perlang.Interpreter/Extensions/TypeExtensions.cs
perlun: Fixed in #228, resolving.

perlun:
First step towards resolving https://github.com/perlang-org/perlang/issues/225.

PR #234, opened 2021-11-22
(interpreter) Fix integer expansions to bigint

perlun:
While working on #225, I discovered that some semantics here was essentially broken. The exponential operator only worked properly with `int` + `int` or `bigint` + `int` operands. When one of the operands was a `float`, it was impossible to assign the result to a variable.

```
$ perlang
Perlang Interactive REPL Console (0.1.0-dev.181, built from git commit 9480f21)
> var v = 2 ** 3.5
[line unknown] Unsupported target type System.Numerics.BigInteger
```

### DIFF
--- a/src/Perlang.Interpreter/PerlangInterpreter.cs
+++ b/src/Perlang.Interpreter/PerlangInterpreter.cs
@@ -8,6 +8,7 @@ using System.Reflection;
 using System.Text;
 using Perlang.Attributes;
 using Perlang.Exceptions;
+using Perlang.Extensions;
 using Perlang.Interpreter.Immutability;
 using Perlang.Interpreter.Internals;
 using Perlang.Interpreter.NameResolution;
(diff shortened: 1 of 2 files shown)


src/Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -1226,9 +1227,22 @@ public VoidObject VisitWhileStmt(Stmt.While stmt)
...
+                    {
+                        int intValue => new BigInteger(intValue),
+                        long longValue => new BigInteger(longValue),
+                        float floatValue => new BigInteger(floatValue),
+                        double doubleValue => new BigInteger(doubleValue),
+
+                        // TODO: Might need to revisit this to support more types as we implement #70.
+                        _ => throw new IllegalStateException($"Unsupported conversion from {value.GetType().ToTypeKeyword()} to {targetTypeReference.ClrType.ToTypeKeyword()}")
+                    };
+                }
```
perlun, line 1242: :point_up: This is the essence of the fix.

Also, the TODO comment for #70 is quite important. The easiest way to ensure that it gets fixed is probably to add more test cases, so that we properly test this with values of all types expected to be supported (i.e. expanding this list as we support more types, `uint`, `ulong` and potentially `decimal` - there are `BigInteger` constructors for all of these that could be used).

src/Perlang.Tests.Integration/Operator/Exponential.cs
```diff
@@ -136,17 +136,19 @@ public void exponential_integer_literal_and_function_return_value()
             Assert.Equal("65536", result);
         }
 
-        [Fact]
-        public void exponential_integer_literals_as_variable_initializer()
+        [Theory]
+        [InlineData("2", "12", "4096")]
+        [InlineData("10", "3.5", "3162")] // TODO: Fix this once the `dynamic` PR gets merged. This is incorrectly coerced to BigInteger in the current implementation.
```
perlun, line 141: :see_no_evil:

perlun:
For reference, the semantics after this MR got merged was **insane**:

```
Perlang Interactive REPL Console (0.1.0-dev.186, built from git commit 94fe2ca)
> var v1 = 2 ** 3.5
> var v2 = 2 ** 3.1
> v1.get_type()
System.Numerics.BigInteger
> v2.get_type()
System.Numerics.BigInteger
> 2 ** 3.5
11.313708498984761
> 2 ** 3.1
8.574187700290345
```

(Using a `BigInteger` type but containing a `double` value?!?)

This was corrected in #237:

```
Perlang Interactive REPL Console (0.1.0-dev.190, built from git commit 3dda89d)
> var v1 = 2 ** 3.5
> var v2 = 2 ** 3.1
> v1.get_type()
System.Double
> v2.get_type()
System.Double
```

PR #237, opened 2021-11-27
(many) Convert binary operators to not use dynamic

perlun:
Closes #257.

### DIFF
--- a/src/Perlang.Interpreter/PerlangInterpreter.cs
+++ b/src/Perlang.Interpreter/PerlangInterpreter.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Globalization;
 using System.Linq;
 using System.Numerics;
 using System.Reflection;
@@ -638,11 +639,11 @@ namespace Perlang.Interpreter
                             return -value;
 
                         case UInt64 value:
-                            // .NET will give an "ambiguous invocation" here, suggesting that we cast value to either a
-                            // `decimal`, `double` or `float`. However, making an integer-type of variable become a
-                            // floating point value all of a sudden seems like odd semantics here. Even though it is
-                            // less performant, we go with the BigInteger approach for now: negating a UInt64 will
-                            // inherently expand it to a BigInteger.
+                            // .NET will give an "ambiguous invocation" here, suggesting that we cast `value` to either
+                            // a `decimal`, `double` or `float`. However, making an integer-type of variable become a
+                            // floating point value all of a sudden seems like very odd and unexpected semantics here.
+                            // Even though it is less performant, we go with the BigInteger approach for now: negating a
+                            // UInt64 will inherently expand it to a BigInteger.
                             return -new BigInteger(value);
 
                         case Single value: // i.e. float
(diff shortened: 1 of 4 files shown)


src/Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -1123,41 +1364,338 @@ public VoidObject VisitWhileStmt(Stmt.While stmt)
...
+                    {
+                        var leftBigInt = (BigInteger)left;
+                        long rightNumber = rightConvertible!.ToInt64(CultureInfo.InvariantCulture);
+
+                        return leftBigInt % rightNumber;
+                    }
+                    else
+                    {
+                        throw new RuntimeError(expr.Operator, $"Operands must be numbers, not {StringifyType(left)} and {StringifyType(right)}");
+                    }
```
perlun, line 1698: :point_up: This (all the changes in this file) was quite a bit of work and I'm not overly pleased with the end result. The problem is that for code that deals with dynamic data of multiple types (`object`), it's really, really hard to do things like this in a clean way.

Possible alternative approach: [use generic math](https://devblogs.microsoft.com/dotnet/preview-features-in-net-6-generic-math/), which is available as a preview in .NET 6. This would possibly _work_ but was nontrivial when I gave it a try. If we would make `Expr.Literal` be a generic type `Expr.Literal<T : INumber>`, it would probably work for constant _numeric_ expressions (but how do we handle literal strings in that case?). We would also need to make it work for `Expr.Identifier`-based expressions as well (i.e. variable references/function calls), and the way our current scanner & parser works, this is simply not doable - the type of an expression is compiled much later than the expression tree is being constructed.

So the TL;DR for generic math is that it's not really helpful for us, AFAIK. (Feel free to prove me wrong, I would really love it if it was doable.)

src/Perlang.Interpreter/Typing/TypeResolver.cs
```diff
@@ -126,18 +126,48 @@ public override VoidObject VisitBinaryExpr(Expr.Binary expr)
...
+                    // Only a certain set of type combinations are supported with this operator. We validate it here to
+                    // be able to provide "compile-time" error checking rather than runtime checking.
+
+                    if ((leftTypeReference.ClrType == typeof(int) || leftTypeReference.ClrType == typeof(long)) &&
+                        rightTypeReference.ClrType == typeof(int))
+                    {
+                        // TODO: This is not 100% sane. The return type depends on the _value_ of the right-hand
+                        // TODO: expression in this case. (See the STAR_STAR implementation in VisitBinaryExpr() for the full
+                        // TODO: details). So it can sometimes be a BigInteger and sometimes a double. This is insane
+                        // TODO: semantics, maybe we should just be consistent and make it always a double in this case...?
```
perlun, line 145: We should perhaps look into this a bit more before merging.

src/Perlang.Parser/Scanner.cs
```diff
@@ -63,6 +63,7 @@ public class Scanner
                 { "byte", RESERVED_WORD },
                 { "sbyte", RESERVED_WORD },
                 { "short", RESERVED_WORD },
+                { "bigint", RESERVED_WORD },
```
perlun, line 66: This is a drive-by fix. We should move this part to a separate PR and add some tests, just like we have for other not-yet-implemented keywords.

perlun: Fixed in #238, resolving.

src/Perlang.Interpreter/Typing/TypeResolver.cs
perlun: Decided to go with the "safer" approach in this case: negative exponents with `int`/`long`/`bigint` values will throw a runtime exception for now, and we retain the `bigint` return type for integer value operands.

At some point, it would be cool to pre-evaluate constant expressions at compile-stage, to be able to make negative constant exponents be a compile-time instead of runtime error. But... for Another Day. :blush:

PR #244, opened 2022-01-10
(ci) Use build matrix for release publishing

perlun:
When it comes to CI, I'm usually a quite big fan of aggressively aiming for parallelizing as much as possible of the build pipeline. I much rather have 5 jobs that run for 30-60 seconds each than one job which runs for 150-300 seconds. It does add a bit of complexity, but I think the benefits typically outweigh that part.

Now that we are changing from 3 supported architecture (`linux-x64`, `osx-x64` and `win-x64`) to five (the previous ones + `linux-arm` and `linux-arm64`), being able to run these builds in parallel does make sense. It also makes the YAML shorter and sweeter, and it makes it simpler to add upcoming new architectures in the future. (`osx-arm64` is
the most likely one to be added quite soon.)

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -21,6 +21,15 @@ jobs:
   publish-packages:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        arch:
+          - linux-arm
+          - linux-arm64
+          - linux-x64
+          - osx-x64
+          - win-x64
+
     steps:
       - uses: actions/checkout@v1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-packages.yml
```diff
@@ -35,105 +44,37 @@ jobs:
...
           SSH_PRIVATE_KEY: ${{ secrets.SSH_PRIVATE_KEY }}
           ARGS: "-rltgoDzvO"
-          SOURCE: "./src/Perlang.ConsoleApp/bin/Release/net6.0/win-x64/*.tar.gz"
+          SOURCE: "./src/Perlang.ConsoleApp/bin/Release/net6.0/${{ matrix.arch }}/*.tar.gz"
           REMOTE_HOST: ${{ secrets.SSH_REMOTE_HOST }}
           REMOTE_USER: ${{ secrets.SSH_REMOTE_USER }}
           TARGET: ${{ secrets.SSH_REMOTE_TARGET }}
         if: github.ref == 'refs/heads/master'
 
+      # TODO: Would ideally run _after_ all jobs of the matrix...
```
perlun, line 77: This is a tricky limitation to get around. There is a `needs` mechanism in GitHub Actions, but it doesn't support propagating all the _files_ (build artifacts) from the jobs to the post-job that way, AFAIK.

I think we'll live with this being executed over and over for now, but the script itself might need to get fixed before we merge this, to avoid creating circular symlinks. (I saw it happening today when running the script manually, with the binaries only for one arch having been uploaded.)

perlun:
Before this change: [2m54s](https://github.com/perlang-org/perlang/actions/runs/1679271680)
After the change: [1m50s](https://github.com/perlang-org/perlang/actions/runs/1679377957) (overall for all the `publish-packages` jobs)

Not perfect in any way, but definitely better at least. :+1:

.github/workflows/publish-packages.yml
perlun: The script was adjusted now. (For reference, here's what it looks like - we should perhaps include this in the `git` repo somehow also - the `rm` call is what got added.)

```shell
#!/bin/bash

get_latest_build() {
    PLATFORM=$1
    LATEST_BUILD=$(ls -1 --sort=time /builds/*-${PLATFORM}.tar.gz | head -n1)

    echo $LATEST_BUILD
}

update_symlink() {
    PLATFORM=$1
    LATEST_BUILD=$2

    ln -sf ${LATEST_BUILD} /builds/perlang-latest-${PLATFORM}.tar.gz
}

# Remove old symlinks, to ensure they don't get included in the get_latest_build() calls
rm /builds/perlang-latest*.tar.gz

update_symlink linux-arm $(get_latest_build "linux-arm")
update_symlink linux-arm64 $(get_latest_build "linux-arm64")
update_symlink linux-x64 $(get_latest_build "linux-x64")
update_symlink osx-x64 $(get_latest_build "osx-x64")
update_symlink win-x64 $(get_latest_build "win-x64")
```

PR #246, opened 2022-01-12
(tests) Improve exceptions in EvalHelper

perlun:
While working on a new language feature, I realized that the way we have been throwing exceptions in `EvalHelper` was not that great. More details in a separate thread.

### DIFF
--- a/src/Perlang.Tests.Integration/EvalHelper.cs
+++ b/src/Perlang.Tests.Integration/EvalHelper.cs
@@ -82,6 +82,40 @@ namespace Perlang.Tests.Integration
         ///
         /// Output printed to the standard output stream will be available in <see cref="EvalResult{T}.Output"/>.
         ///
+        /// This method will propagate all errors apart from  <see cref="ScanError"/> to the caller. Scan errors
+        /// will be available in the returned <see cref="EvalResult{T}.Errors"/> property.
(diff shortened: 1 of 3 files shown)


src/Perlang.Tests.Integration/EvalHelper.cs
```diff
@@ -268,32 +268,32 @@ internal static object EvalWithArguments(string source, params string[] argument
...
         private static void AssertFailValidationErrorHandler(ValidationError validationError)
         {
-            throw validationError;
+            throw new Exception("ValidationError occurred. See inner exception for details.", validationError);
         }
 
         private static bool AssertFailCompilerWarningHandler(CompilerWarning compilerWarning)
         {
-            throw compilerWarning;
+            throw new Exception("CompilerWarning occurred. See inner exception for details.", compilerWarning);
```
perlun, line 296: The previous approach (creating an exception in one place, throwing it elsewhere) had one major drawback: it completely destroys the debugging experience, since the stack trace gets generated _when the exception is thrown_. Hence, it's much better to use the `InnerException` approach for this.

(#232 was similar to this, but the problem at that time was re-throwing new exceptions without providing the inner exception details; they were silently discarded, also negatively affecting the debugging experience.)

Issue #245, opened 2022-01-11
Add boolean operators


perlun:
We currently don't support the standard boolean operators `&&` and `||`. We should implement these, and give them the same semantics as in the C language family (C, Java, C#) with _one important difference_:

Doing things which are considered "hard-to-from" from an ambiguity should produce a compile-time error. I am thinking about expressions like this:

`true && false || false`

Without digging up a book/Internet resource, can you _by heart_ say what the above will produce in the language you use everyday?

(The right answer is typically _`false`_ since `||` has higher precedence than `&&`. In other words, the above should be read as `true && (false || false)`.)

### What Perlang aims to do here

We want to make it _impossible_ to write the above, ambiguous code. The compiler should complain and enforce the usage of clarifying parentheses in this case.

#### Examples of code that should cause a compilation error

`true && false || false`

#### Examples of code that should work

```perlang
true && (false || false)
(true && false) || false
```

perlun:
This is now mostly in place. :tada: What remains is to implement the checks mentioned here:

> ### What Perlang aims to do here
> 
> We want to make it _impossible_ to write the above, ambiguous code. The compiler should complain and enforce the usage of clarifying parentheses in this case.
>
> #### Examples of code that should cause a compilation error
> 
> `true && false || false`
>
> #### Examples of code that should work
> 
> ```
> true && (false || false)
> (true && false) || false
> ```

PR #247, opened 2022-01-14
(many) Implement shift left and shift right operators

perlun:


src/Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -1702,6 +1703,69 @@ public VoidObject VisitWhileStmt(Stmt.While stmt)
                     }
                     else
                     {
+                        // TODO: "Operands must be numbers" isn't completely correct here.
```
perlun, line 1706: ...since the operands can be numbers, but of unsupported types. We'll live with this being slightly incorrect for now.

src/Perlang.Tests.Integration/Operator/ShiftOperators.cs
```diff
@@ -0,0 +1,201 @@
...
+            [Fact]
+            public void takes_precedence_over_power_operator()
+            {
+                string source = @"
+                    print 1024 >> 8 ** 10;
+                ";
+
+                string result = EvalReturningOutputString(source);
+                Assert.Equal("1048576", result);
+            }
```
perlun, line 184: Some of these like the `%` and `**` examples are arguably a bit weird. _Why would anyone in their right mind want to do something like that?_ On the other hand, these examples are not outright _dangerous, as is the case described in #245 (with `&&` and `||`). As long as we keep the precedence of `<<` and `>>` very high, like we currently do, we should be pretty safe.

While the above example is unlikely, I consider the fact that the user would actually want to do `1024 >> (8 ** 10)` even weirder... :slightly_smiling_face: 

Anyway, we'll live with this being a valid combination of operators now, until proven otherwise by experience.

PR #252, opened 2022-01-16
(tests) #46 Add `and` tests from Lox

perlun:


src/Perlang.Common/Extensions/TypeExtensions.cs
```diff
@@ -18,6 +18,9 @@ public static string ToTypeKeyword(this Type type)
                 { } when type == typeof(BigInteger) => "bigint",
                 { } when type == typeof(NullObject) => "null",
                 { } when type == typeof(String) => "string",
+
+                // TODO: add bool here
+
```
perlun, line 23: ...in another PR.

src/Perlang.Tests.Integration/LogicalOperator/And.cs
```diff
@@ -0,0 +1,159 @@
...
+        public void returns_the_last_argument_if_all_are_truthy_1_and_2_and_3()
+        {
+            string source = @"
+                print 1 and 2 and 3;
+            ";
+
+            string output = EvalReturningOutput(source).SingleOrDefault();
+
+            Assert.Equal("3", output);
+        }
```
perlun, line 70: This whole "returns the last argument" stuff is just sketchy all the way from top to bottom. It may fit in  dynamic languages like Lox or Ruby (the latter has the similar behavior _but only with `&&`_, not with its different-precedence-operator named `and` :see_no_evil: :hear_no_evil: :speak_no_evil:) but it definitely feels odd in a Perlang context. We should revamp this before we consider #245 "done".

src/Perlang.Tests.Integration/LogicalOperator/And.cs
```diff
@@ -0,0 +1,159 @@
...
+        public void empty_string_is_truthy()
+        {
+            string source = @"
+                print """" and ""ok"";
+            ";
+
+            string output = EvalReturningOutput(source).SingleOrDefault();
+
+            Assert.Equal("ok", output);
+        }
```
perlun, line 157: Likewise, this whole "non-bool is truthy/falsy" fits very poorly in the Perlang context, so it should also be fixed sooner or later.

PR #255, opened 2022-01-18
(interpreter) Make boolean operators and keywords perform type-checking

perlun:
This commit addresses the following issues:

- Lox had a fairly relaxed (or if you will, non-conforming) way of handling boolean operators like `and` and `or`. (These operators were renamed to `&&` and `||` in #253, to make the language feel more like our friends in C# and Java.)

  The way these operators worked was to try and coalesce pretty much everything to a "truthy" or "falsy" value, much like languages like Ruby and JavaScript operate. This makes sense in a dynamic language context, but Perlang strives to have a rich type checker and we consider static typing a feature. Hence, it makes sense to be more explicit when it comes to boolean operator operands.

- The result from the `and` and `or` operators also followed a certain pattern, described in detail here:  https://craftinginterpreters.com/control-flow.html#logical-operators. We want these methods to always return a `bool` value. Hence, the `TypeResolver` is amended in this commit to make the CLR type of all `Expr.Logical` instances be `bool`.

The end result is something closer to the semantics we are aiming for in issue #245.

Related ticket: #245
Related ticket: #43

### DIFF
--- a/src/Perlang.Interpreter/PerlangInterpreter.cs
+++ b/src/Perlang.Interpreter/PerlangInterpreter.cs
@@ -569,30 +569,43 @@ namespace Perlang.Interpreter
             return expr.Value;
         }
 
-        public object? VisitLogicalExpr(Expr.Logical expr)
+        public object VisitLogicalExpr(Expr.Logical expr)
         {
-            object? left = Evaluate(expr.Left);
+            // Note: ! operators in this method are considered "safe" since the type validation is expected to have
+            // validated that these operators are only called with `bool` operands.
+
+            object left = Evaluate(expr.Left)!;
 
             if (expr.Operator.Type == PIPE_PIPE)
             {
-                if (IsTruthy(left))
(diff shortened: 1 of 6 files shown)


src/Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -569,30 +569,43 @@ private void Interpret(IEnumerable<Stmt> statements)
             return expr.Value;
         }
 
-        public object? VisitLogicalExpr(Expr.Logical expr)
+        public object VisitLogicalExpr(Expr.Logical expr)
         {
-            object? left = Evaluate(expr.Left);
+            // Note: ! operators in this method are considered "safe" since the type validation is expected to have
+            // validated that these operators are only called with `bool` operands.
```
perlun, line 575: I'm not 100% sure this is a good idea or not... It essentially makes some parts of the code blindly trust other parts of the code. This _can_ be a horrible idea, or perhaps it's totally genius. Time will tell. :slightly_smiling_face:

PR #257, opened 2022-01-22
(ci) Run tests on all supported platforms

perlun:
It is likely that we will soon introduce some functionality which is OS-dependent (e.g. making it possible to support a subset of the POSIX system calls); these changes are unlikely to work on Windows. To ensure a consistent (or at least _predictable_) experience on all platforms, it's important that we consistently run all tests on all supported platforms from now on.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,27 +1,17 @@
-name: .NET Core build & test
+name: .NET run tests
 
 on: [push]
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v1
-
-      - name: Setup .NET
-        uses: actions/setup-dotnet@v1
-        with:
-          dotnet-version: 6.0.x
-
-      - name: Re-generate auto-generated files
-        run: make auto-generated
-
-      - name: Build with dotnet
-        run: dotnet build --configuration Release
-
   test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
+
+    strategy:
+      matrix:
+        runner:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yml
```diff
@@ -4,7 +4,14 @@ on: [push]
...
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
+
+    strategy:
+      matrix:
+        runner:
+          - windows-2022
+          - ubuntu-20.04
+          - macos-11
```
perlun, line 14: (Hmm, I wonder if we really need to _build_ on all supported platforms like this. There is probably a bit of redundancy in that we test the `publish-packages` job - on Linux - anyway, so maybe we could leave this out altogether. Will think about it, it will let us skip some of the CI jobs if we make that change.)

.github/workflows/build-test.yml
```diff
@@ -21,7 +28,14 @@ jobs:
...
   test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
+
+    strategy:
+      matrix:
+        runner:
+          - windows-2022
+          - ubuntu-20.04
+          - macos-11
```
perlun, line 38: The tests are currently failing on Windows; will look into it separately.

.github/workflows/build-test.yml
perlun: Removed the `build` job. We'll let `publish-packages` take care of this and minimize the duplication this way.

.github/workflows/build-test.yml
perlun: Fixed in #258.

PR #258, opened 2022-01-22
(tests) Make ProgramTest work on Windows

perlun:
Also, convert this test to Fluent Assertions.

### DIFF
--- a/src/Perlang.Tests/ConsoleApp/ProgramTest.cs
+++ b/src/Perlang.Tests/ConsoleApp/ProgramTest.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.CommandLine.IO;
 using System.Linq;
+using FluentAssertions;
 using Perlang.ConsoleApp;
 using Perlang.Parser;
 using Xunit;
@@ -36,7 +37,7 @@ namespace Perlang.Tests.ConsoleApp
             {
                 subject.Run("var a = 42; print a;", FatalWarningsHandler);
 
-                Assert.Equal(new List<string> { "42" }, output);
+                output.Should().Equal("42");
             }
 
             [Fact]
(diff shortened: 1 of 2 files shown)


perlun:
Initially includes #257 until CI has executed; I need the Windows CI test to validate this change before merging it. Once this is is in `master`, we should then be able to rebase #257 and get a clean CI run on that one as well. :tada:

perlun:
> Initially includes #257 until CI has executed; I need the Windows CI test to validate this change before merging it.

Seemed to work fine. Auto-merge enabled now.

src/Perlang.Tests/ConsoleApp/ProgramTest.cs
```diff
@@ -81,7 +82,7 @@ public void can_call_function_from_expression()
                 subject.Run("fun hello(): void { print 1; }", FatalWarningsHandler);
                 subject.Run("hello()", FatalWarningsHandler);
 
-                Assert.Equal(new List<string> { "1" }, output);
+                output.Should().Equal("1");
```
perlun, line 85: For cases like these, Fluent Assertions feels quite obviously like a "better" approach. For others, maybe not so much. Anyway, I think it does make sense in another way as well in that it makes us less dependent on xUnit. If we ever want to switch to some other unit testing framework, we are much easier off if we're already using Fluent Assertions for our assertions => gain.

src/Perlang.Tests/ConsoleApp/ProgramTest.cs
```diff
@@ -172,31 +181,46 @@ public class WithPrintParameter
...
+                /// This property returns the output split by newline separators.
+                /// </summary>
+                private IEnumerable<string> StdoutLines => StdoutContent.Split(Environment.NewLine);
+
                 [Fact]
                 public void assignment_and_increment()
                 {
                     CallWithPrintParameter("i = i + 1");
 
-                    Assert.Equal("(i (+ i 1))\n", StdoutContent);
```
perlun, line 183: This was the problem this MR fixed: there were hardwired assumptions on `Environment.NewLine` being `\n` (LF), which is the case on Linux and macOS but definitely not on Windows. By converting this to checking the `StdoutLines` (added in this MR), we both fixed this and made it more convenient to write these assertions at the same time (since we can now check an `IEnumerable` of `string`s instead of a long `string`).

PR #259, opened 2022-01-23
(tests) Improve exception messages from eval

perlun:


src/Perlang.Tests.Integration/EvalHelper.cs
```diff
@@ -303,32 +303,34 @@ internal static object EvalWithArguments(string source, params string[] argument
...
-            throw new Exception("ScanError occurred. See inner exception for details.", scanError);
+            // Note: does not contain stack trace, so no point in wrapping in Exception() at this point.
+            throw scanError;
         }
 
         private static void AssertFailParseErrorHandler(ParseError parseError)
         {
-            throw new Exception("ParseError occurred. See inner exception for details.", parseError);
+            // Note: does not contain stack trace, so no point in wrapping in Exception() at this point.
+            throw parseError;
```
perlun, line 313: These hunks revert part of #246; it doesn't add any value to wrap these exceptions, since they haven't been thrown yet in the first place => no stack trace has been added to the `ScanError`/`ParseError` objects. We might as well just keep it like it used to be for now.

src/Perlang.Tests.Integration/EvalHelper.cs
```diff
@@ -303,32 +303,34 @@ internal static object EvalWithArguments(string source, params string[] argument
...
         private static void AssertFailValidationErrorHandler(ValidationError validationError)
         {
-            throw new Exception("ValidationError occurred. See inner exception for details.", validationError);
+            throw new Exception($"ValidationError occurred: {validationError.Message}", validationError);
         }
 
         private static bool AssertFailCompilerWarningHandler(CompilerWarning compilerWarning)
         {
-            throw new Exception("CompilerWarning occurred. See inner exception for details.", compilerWarning);
+            throw new Exception($"CompilerWarning occurred: {compilerWarning.Message}", compilerWarning);
```
perlun, line 333: These hunks makes the inner exception message be immediately apparent in e.g. Rider's "last test result" overlay when editing code, which is nice since it makes it super-easy to see why a particular test is failing while you are working on it:

image

PR #260, opened 2022-01-23
(stdlib) Add Posix module with some initial methods

perlun:


perlun:
This seems to work quite nicely, and we can easily provide some of the basic `get*` methods here, providing a smooth experience for people writing Perlang code on POSIX-based systems like Linux and macOS. We probably need to figure out some reasonable way to gracefully fail on Windows, though. (since it doesn't support this, or...?)

We added Windows CI in #257; this will help us learn how this behaves on non-POSIX platforms.

perlun:
> We probably need to figure out some reasonable way to gracefully fail on Windows, though. (since it doesn't support this, or...?)

Added in #266. The approach taken there is certainly not _perfect_ but it will do for now.

Directory.Build.props
```diff
@@ -6,6 +6,18 @@
...
+    </PropertyGroup>
+
+    <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('OSX'))">
+        <DefineConstants>_MACOS</DefineConstants>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('Windows'))">
+        <DefineConstants>_WINDOWS</DefineConstants>
+    </PropertyGroup>
+
```
perlun, line 20: Inspired by https://stackoverflow.com/a/65208710/227779 and https://blog.magnusmontin.net/2018/11/05/platform-conditional-compilation-in-net-core/, this approach seems to be a quite nice way to handle platform-specific code. :tada:

Perlang.sln.DotSettings
```diff
@@ -9,6 +9,10 @@
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpUseContinuousIndentInsideBracesMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateBlankLinesAroundFieldToBlankLinesAroundProperty/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Callables/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=getcwd/@EntryIndexedValue">True</s:Boolean>
```
perlun, line 12: (I initially planned to have a `getcwd` which delegated to `Environment.CurrentDirectory`, but I concluded it wasn't a great idea since `getcwd` actually _exists_ on Windows as well... I am leaning towards adding a `Libc` or `System` class or something where we can put stuff like this. In a way, `Libc` would perhaps make most sense if we consider it a container for "C-like" functions like `getcwd`, perhaps `fopen` (will be needed for e.g. `Posix.mmap` I think) and so forth.)

PR #261, opened 2022-01-24
(many) Make number-to-string conversions use invariant culture

perlun:
I discovered these errors by accident, when trying to run the test suite on Windows (for an _entirely_ different purpose 🙂). We had some assumptions in the unit tests which failed because the `ToString()` conversions in .NET uses the `CultureInfo.CurrentCulture` by default. This means that if you have your Windows running with e.g. Swedish regional settings (which is my setup), a `double` variable with a value of `123.45` will be rendered as `123,45`.

While this is preferable in some contexts, I think we'll go with the "safe and consistent" approach for now, if not for any other reason that the unit/integration tests currently depend on these semantics. It doesn't provide a good dev experience for people trying this out on Windows/with their machine configured to use a different locale, if they would get a bunch of silly errors.


src/Perlang.Common/Utils.cs
```diff
@@ -13,6 +15,12 @@ public static string Stringify(object @object)
             {
                 return "null";
             }
+            else if (@object is IConvertible convertible)
+            {
+                // The explicit IFormatProvider is required to ensure we use 123.45 format, regardless of
+                // host OS language/region settings. See #xxx for more details.
```
perlun, line 21: **TODO**: create a proper issue for this and amend these comments, pointing the reader to the issue in question.

perlun:
(Temporarily cherry-picked the changes from #262 into this branch to ensure that it runs cleanly before merging. We'll need to merge this _first_ and then the other one, because this fixes the underlying problems that the CI PR highlights.)

perlun:
> (Temporarily cherry-picked the changes from #262 into this branch to ensure that it runs cleanly before merging. We'll need to merge this _first_ and then the other one, because this fixes the underlying problems that the CI PR highlights.)

CI is clean. Will fix some remaining issues in this and get it merged, so we can get #262 to also be mergeable (CI clean).

src/Perlang.Common/Utils.cs
perlun: #263

PR #262, opened 2022-01-25
(ci) Run tests on multiple locales

perlun:
The aim with these tests is to prevent regressions/bugs which are only present when a particular locale is used; there are currently some such bugs in our current `master` codebase.


perlun:
Closed in favor of #264.

PR #264, opened 2022-01-26
(tests) Run locale-dependent tests on multiple locales

perlun:
This provides an alternative approach to #262. Instead of running _the whole test suite_ on multiple locales, we run the particular tests which was seen to fail before #261 got merged, on those locales.

(For the curious ones, the way this was made was to make a temporary commit & branch on top of cdaed13, i.e. the parent commit of https://github.com/perlang-org/perlang/commit/b384c74bf20d7d0c9f42d94d6f8b0d97ebb7f370 / #261. That way, I could reproduce the previous failures quite easily to be able to tailor-fit these locale-specific changes by switching from `[Fact]` to `[Theory]`/data-driven tests for these test methods only.)

This does involve a certain cost in terms of more test code and slightly added complexity in that regard. The pros:

- _Slightly simpler CI config_
- _Potentially slightly faster CI_ (I haven't timed this, but we at least have 1 less job to run for each PR)
- _Can customize more as needed_. If we ever run into some edge cases which are only failing on a particular locale (God forbid), we could potentially add more parameterized tests for _that particular test only_, which could be useful.

### DIFF
--- a/src/Perlang.Tests.Integration/Number/NumberTests.cs
+++ b/src/Perlang.Tests.Integration/Number/NumberTests.cs
@@ -1,4 +1,6 @@
+using System.Globalization;
 using System.Linq;
+using System.Threading.Tasks;
 using Xunit;
 using static Perlang.Tests.Integration.EvalHelper;
 
@@ -95,9 +97,12 @@ namespace Perlang.Tests.Integration.Number
             Assert.Equal(-123, result);
         }
 
-        [Fact]
-        public void literal_float()
+        [Theory]
+        [ClassData(typeof(TestCultures))]
+        public async Task literal_float(CultureInfo cultureInfo)
         {
+            CultureInfo.CurrentCulture = cultureInfo;
+
             string source = @"
                 123.456
             ";
@@ -107,9 +112,12 @@ namespace Perlang.Tests.Integration.Number
             Assert.Equal(123.456, result);
         }
 
-        [Fact]
-        public void literal_negative_float()
+        [Theory]
+        [ClassData(typeof(TestCultures))]
+        public async Task literal_negative_float(CultureInfo cultureInfo)
         {
+            CultureInfo.CurrentCulture = cultureInfo;
+
             string source = @"
                 -0.001
             ";
@@ -119,9 +127,12 @@ namespace Perlang.Tests.Integration.Number
             Assert.Equal(-0.001, result);
         }
 
-        [Fact]
-        public void literal_float_with_underscore_in_integer_part()
+        [Theory]
+        [ClassData(typeof(TestCultures))]
+        public async Task literal_float_with_underscore_in_integer_part(CultureInfo cultureInfo)
         {
+            CultureInfo.CurrentCulture = cultureInfo;
+
             string source = @"
                 123_45.678
             ";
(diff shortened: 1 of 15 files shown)


src/Perlang.Tests.Integration/Number/NumberTests.cs
```diff
@@ -95,9 +97,12 @@ public void literal_negative_integer()
...
         }
 
-        [Fact]
-        public void literal_float()
+        [Theory]
+        [ClassData(typeof(TestCultures))]
+        public async Task literal_float(CultureInfo cultureInfo)
         {
+            CultureInfo.CurrentCulture = cultureInfo;
+
```
perlun, line 105: An important aspect regarding this is that changing the `CurrentCulture` _without making the test method be `async Task`_ would be dangerous, since the culture change would leak over to other test methods as well. Given this, I think we should add an ArchUnit.NET test to ensure that all test methods which take a `CultureInfo` parameter **must** be `async` and return a `Task`.

(The latter check is in fact not mandatory to check, since there's already an inspection in place that will cause errors on `async void` methods, which is definitely a dangerous anti-pattern that should be avoided.)

src/Perlang.Tests.Integration/Operator/Comparison.cs
```diff
@@ -261,9 +263,12 @@ public void zero_and_negative_zero_integers_are_identical()
             Assert.Equal(false, output);
         }
 
-        [Fact]
-        public void zero_less_than_negative_zero_is_false()
+        [Theory]
+        [ClassData(typeof(TestCultures))]
+        public void zero_less_than_negative_zero_is_false(CultureInfo cultureInfo)
```
perlun, line 268: Oops, this illustrates the problem mentioned in https://github.com/perlang-org/perlang/pull/264/files#r793446103. :see_no_evil: It's very easy to make these kind of mistakes by accident.

src/Perlang.Tests.Integration/Operator/Exponential.cs
```diff
@@ -86,9 +88,12 @@ public void exponential_negative_and_positive_integer_literals_again()
             Assert.Equal(new BigInteger(59049), result);
         }
 
-        [Fact]
-        public void exponential_integer_and_float_literals()
+        [Theory]
+        [ClassData(typeof(TestCultures))]
+        public void exponential_integer_and_float_literals(CultureInfo cultureInfo)
```
perlun, line 93: https://github.com/perlang-org/perlang/pull/264/files#r793446103

src/Perlang.Tests.Integration/ReservedKeywords/ReservedKeywordsTests.cs
```diff
@@ -485,7 +485,7 @@ public void function_return_type_detects_reserved_keyword_char()
         {
             string source = @"
                 fun foo(): char {
-                    return 123.45;
+                    return 123;
```
perlun, line 488: These are just simplifications to avoid making these tests use constructs which can have locale-dependent bugs. (the kind of things fixed in #261)

src/Perlang.Tests.Integration/Stdlib/Base64EncodeTests.cs
```diff
@@ -45,12 +45,12 @@ public void Base64_encode_with_a_long_string_argument_returns_the_expected_resul
...
         public void Base64_encode_with_a_numeric_argument_throws_the_expected_exception()
         {
-            var result = EvalWithValidationErrorCatch("Base64.encode(123.45)");
+            var result = EvalWithValidationErrorCatch("Base64.encode(123)");
             var runtimeError = result.Errors.First();
 
             Assert.Single(result.Errors);
 
-            Assert.Equal("Cannot pass double argument as string parameter to encode()", runtimeError.Message);
+            Assert.Equal("Cannot pass int argument as string parameter to encode()", runtimeError.Message);
```
perlun, line 53: Likewise, using `int` here is simpler since we don't need to demonstrate the locale-dependent bugs in this context. (they are already tested elsewhere)

PR #266, opened 2022-01-29
(interpreter) Support platform-specific global classes

perlun:
We will likely never get full POSIX support on Winodws; this is simply a fact that we'll have to live with. Different languages and runtimes take different approaches to this:

- Java and .NET try to abstract away the fact that operating systems   differ, to various degrees. Java sometimes takes this to rather absurd lengths, by even making it impossible to do simple things like calling  `getuid`. "Write once, run anywhere" definitely comes at a price.

- .NET have traditionally done it slightly differently, given that it   was originally very tied to Windows. Relying heavily on Windows-only APIs like GDI+ and DirectX was not a problem in this context. Now that .NET is cross-platform, things are indeed a bit different and perhaps moving slightly more towards Java (while still providing Windows-specifics _for apps that need it_).

Perlang is likely to land more close to the .NET position on this. I really like the idea of providing a `Posix` module/class, with things like `getgid`, `getuid`, `getpid` and so forth readily available. When we support imports, I even imagine being able to do something like `import Posix.*` or `import static Posix.*`, to be able to call these methods pretty much exactly the same as in C.

Imagine this for a moment. Something that "feels like C" (in a positive sense), but still can run in a managed context, even in interpreted mode, doesn't that sound pretty awesome to you? (Someone will probably soon start saying "you're reinventing Python, you know that?" to me soon; I will do my very best to not pay any attention to those voices at all. :grin:)

### DIFF
--- a/src/Perlang.Common/Attributes/GlobalClassAttribute.cs
+++ b/src/Perlang.Common/Attributes/GlobalClassAttribute.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using JetBrains.Annotations;
 
 namespace Perlang.Attributes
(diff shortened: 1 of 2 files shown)


src/Perlang.Common/Attributes/GlobalClassAttribute.cs
```diff
@@ -23,5 +24,22 @@ public class GlobalClassAttribute : Attribute
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="GlobalClassAttribute"/> class.
+        /// </summary>
+        /// <param name="platforms">An optional list of platform IDs for which the given global class should be
+        /// registered. If no platform IDs are provided, the global class is presumed to be platform-agnostic and
+        /// registered on all supported platforms.</param>
+        public GlobalClassAttribute(params PlatformID[] platforms)
+        {
+            Platforms = platforms;
+        }
```
perlun, line 43: Note that this approach has one major advantage: the code for a particular unsupported platform will still be _included_ (although not _active_) on that platform. This might not be a huge problem at first, but eventually we might want to be able to exclude it from the `.dll` altogether.

One way around it which would be possible right now would be to use `#if` directives to conditionally exclude this on particular platforms. This has the disadvantage of making it hard/impossible to work on this code on the unsupported platform though (the IDE will show this part of the code as "excluded"), so... not wanting to go that route for this particular reason.

PR #269, opened 2022-02-05
(stdlib) Add Libc class with some basic functions

perlun:
Also moved one function from `Posix` to `Libc`, since it's available on all supported platforms. :tada:

### DIFF
--- a/src/Perlang.Stdlib/Stdlib/Libc.cs
+++ b/src/Perlang.Stdlib/Stdlib/Libc.cs
@@ -1,0 +1,178 @@
+#nullable enable
+#pragma warning disable SA1300
+#pragma warning disable S3218
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.ComponentModel;
+using System.Runtime.InteropServices;
+using Perlang.Attributes;
+
+namespace Perlang.Stdlib
+{
+    /// <summary>
+    /// Provides support for calling standard C library functions.
+    ///
+    /// This class supplements see <see cref="Posix"/> class by implementing support for functions available on all
+    /// supported platforms.
+    /// </summary>
+    /// <remarks>
+    /// The XML method descriptions are based on [the .NET source code](https://github.com/dotnet/runtime), licensed
+    /// under the MIT license. Copyright (c) .NET Foundation and Contributors. Some method descriptions are also based
+    /// on `man` pages in [the NetBSD source code](https://github.com/NetBSD/src). The full license of these man pages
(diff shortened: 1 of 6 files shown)


src/Perlang.Stdlib/Stdlib/Posix.cs
```diff
@@ -1,62 +1,44 @@
...
-    ///    notice, this list of conditions and the following disclaimer.
-    /// 2. Redistributions in binary form must reproduce the above copyright
-    ///    notice, this list of conditions and the following disclaimer in the
-    ///    documentation and/or other materials provided with the distribution.
-    /// 3. Neither the name of the University nor the names of its contributors
-    ///    may be used to endorse or promote products derived from this software
-    ///    without specific prior written permission.
+    /// * <see cref="Libc.getcwd"/>
+    /// * <see cref="Libc.getenv"/>
+    /// * <see cref="Libc.getpid"/>
```
perlun, line 26: It may end up being unmanageable to list all `Libc` methods here like this, but... let's give it a try. :+1:

src/Perlang.Stdlib/Stdlib/Libc.cs
```diff
@@ -0,0 +1,109 @@
...
+        ///
+        /// On Windows systems, the environment block of the current process includes:
+        ///
+        /// - All environment variables that are provided to it by the parent process that created it. For example, a
+        ///   .NET application launched from a console window inherits all of the console window's environment
+        ///   variables.
+        ///
+        ///       If there is no parent process, per-machine and per-user environment variables are used instead. For
+        ///       example, a new console window has all per-machine and per-user environment variables defined at the
+        ///       time it was launched.
```
perlun, line 74: The formatting here is indeed a bit "off", but the weird thing is that this is _required_. By indenting this as "ordinary Markdown", the content doesn't continue the previous paragraph. :neutral_face: Really weird. I guess we should `markdig` a shot, to see if it behaves the same or differently.

src/Perlang.Tests.Integration/Stdlib/LibcTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        // TODO: This can be improved when #270 is implemented.
+        [Fact]
+        public void environ_item_supports_get_Item()
+        {
+            var result = Eval("Libc.environ().get_Item(\"PATH\")");
+
+            result.Should()
+                .BeOfType<string>().Which.Should()
+                .NotBeEmpty();
+        }
```
perlun, line 42: This test illustrated a deficiency in the language. See #270 for more details.

PR #280, opened 2022-02-17
(docs) Add 'learn' section

perlun:
I've been working on this for far too long. Not _actively_ working of course, but still... writing a bit of content here and there. In fact, I started it at this time:

> Fri Apr 16 22:49:11 2021 +0300

Almost a full _year_ for a single branch. That's way too long. But then again, writing (good!) documentation is really really hard. At least to me, it feels like it takes energy on a _completely different level_ than writing software. (This probably correlates strongly to the fact that I've spent way more time on writing software than writing docs in my years.)

Oh well, time to get it merged & published now. :tada:


docs/about/index.md
```diff
@@ -2,4 +2,4 @@
 
 In these pages, we try to describe both what Perlang _is_ and what it aims to _become_. We try to make the distinction betwen these clear, when relevant.
 
-If you just want to jump straight into some examples so you can get a feeling of what it looks like, head straight to the [learn](learn) subpage.
+If you just want to jump straight into some examples so you can get a feeling of what it looks like, head straight to the [learn](../learn/quickstart/index.md) section.
```
perlun, line 5: These kind of changes are a bit unpleasant since they will include `index.html` in the URLs (which I think looks very ugly, TBH), but it unfortunately plays better with DocFx. (It emits a warning for links that do not specifically point at a particular page - if anyone reading this knows how to work around this, patches/suggestions are highly appreciated!)

PR #281, opened 2022-02-18
(stdlib) Minor docs improvement

perlun:


src/Perlang.Stdlib/Stdlib/Libc.cs
```diff
@@ -64,6 +64,17 @@ public static string getcwd()
...
+        /// ### On Linux and macOS systems
+        ///
+        /// On Linux and macOS, the <see cref="environ"/> method retrieves the name and value of all environment
+        /// variables that are inherited from the parent process that launched the `dotnet` process or that are defined
+        /// within the scope of the `dotnet` process itself. Once the `dotnet` process ends, these latter environment
+        /// variables cease to exist.
+        ///
+        /// .NET running on Unix-based systems does not support per-machine or per-user environment variables.
+        ///
+        /// ### On Windows systems
```
perlun, line 76: This is the header that was missing that triggered this change. I also though it made sense to invert the order here (placing Linux and macOS first), since Perlang is developed primarily on Linux. (it definitely _works_ on Windows as well but I suspect that it will have some Unix-only features over time, more so than "Windows-only" functionality)

This is fundamentally different from .NET, which grew out of .NET Framework (which was Windows-only) and only started adding full Unix support from 2014 and onwards.

PR #282, opened 2022-02-19
(scripts) Change -dev tag to have v prefix

perlun:
This is a preparational step for making the 0.1.0 release. Upon careful consideration and taking e.g. [this SO thread](https://stackoverflow.com/questions/21639437/git-flow-release-branches-and-tags-with-or-without-v-prefix) into consideration, I've decided to name the 0.1.0 tag `v0.1.0`. The argument about being able to tab-complete them, and also differentiate between e.g. `0.1` release branches easily, does make a certain bit of sense.

Doing this requires some changes in the way we generate the `src/Perlang.Common/CommonConstants.Generated.cs` file though, which is covered by this PR.

### DIFF
--- a/src/Perlang.Stdlib/Stdlib/Libc.cs
+++ b/src/Perlang.Stdlib/Stdlib/Libc.cs
@@ -168,7 +168,7 @@ namespace Perlang.Stdlib
         /// to use it for constructing temporary file names. However, this is _NOT_ recommended since it is not secure.
         /// In C, `mkstemp` can be used for this purpose. `mkstemp` is not yet available to use from Perlang, though.
         ///
-        /// <see cref="getpid">getpid()</see> conform to ISO/IEC 9945-1:1990 (`POSIX.1`).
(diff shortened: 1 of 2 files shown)


src/Perlang.Stdlib/Stdlib/Libc.cs
```diff
@@ -168,7 +168,7 @@ public static string getcwd()
         /// to use it for constructing temporary file names. However, this is _NOT_ recommended since it is not secure.
         /// In C, `mkstemp` can be used for this purpose. `mkstemp` is not yet available to use from Perlang, though.
         ///
-        /// <see cref="getpid">getpid()</see> conform to ISO/IEC 9945-1:1990 (`POSIX.1`).
+        /// <see cref="getpid">getpid()</see> conforms to ISO/IEC 9945-1:1990 (`POSIX.1`).
```
perlun, line 171: (Drive-by fix, fixing a typo I spotted yesterday.)

PR #284, opened 2022-02-19
(consoleapp) Add /help command

### LINKED ISSUE #212: Add /help command in REPL
The current way the REPL works is not very helpful. You don't get a whole lot of information from it on how it should be used; it doesn't even tell you how to exit it! (Now, that's "semi-easy" to guess since both `quit` and `exit` works, but still...)

We should add a basic `/help` command which displays something like the following:

- `/quit` and `/exit`, and potentially other internal REPL commands
- Some examples of statements/expressions that works in the REPL
- Preferably, a pointer (URL) to some tutorial-oriented documentation that can be kept always-up-to-date.

perlun:
Closes #212.

### DIFF
--- a/src/Perlang.ConsoleApp/Program.cs
+++ b/src/Perlang.ConsoleApp/Program.cs
@@ -354,9 +354,14 @@ namespace Perlang.ConsoleApp
             {
                 string command = ReadLine.Read("> ");
 
-                if (command == "quit" || command == "exit")
+                if (command.ToLowerInvariant() == "/quit" || command.ToLowerInvariant() == "/exit")
                 {
                     break;
+                }
+                else if (command.ToLowerInvariant() == "/help")
+                {
+                    ShowHelp();
+                    continue;
                 }
 
                 // REPL mode is more relaxed: -Wno-error is enabled for all warnings by default. It simply makes sense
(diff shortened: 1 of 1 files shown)


src/Perlang.ConsoleApp/Program.cs
```diff
@@ -388,7 +416,9 @@ private void ParseAndPrint(string source)
 
         private void PrintBanner()
         {
-            standardOutputHandler($"Perlang Interactive REPL Console ({CommonConstants.GetFullVersion()}, built from git commit {CommonConstants.GitCommit})");
+            standardOutputHandler($"Perlang Interactive REPL Console (\x1B[1m{CommonConstants.GetFullVersion()}\x1B[0m, built from git commit {CommonConstants.GitCommit})");
+            standardOutputHandler("Type \x1B[36;1m/help\x1B[0m for more information or \x1B[36;1m/quit\x1B[0m to quit the program.");
+            standardOutputHandler(String.Empty);
```
perlun, line 421: The careful reader will notice that this makes quite heavy use of ANSI colouring sequences. This should even work on Windows nowadays, as of Windows 10 version 1909 (released in November 2019). I haven't verified it, but will test it at some point after merging this.

The REPL is always expected to be executed in "interactive" mode, but we should probably try to auto-detect the support for colour at some point. Or at the very least, add a `--[no-]color` option to be able to turn it on or off.

(Side note, but I know I might seem completely inconsistent with the "color" vs "colour" usage above. but I'm not sure I really am. I think that `--color` makes more sense as a command line option, since more people are likely to expect the Americanized spelling of this word. At the same time, I think I learned the British spellings in school (in Sweden where I grew up), and I kind of like it better for some odd reason. It feels more colourful or something... :smirk:)

src/Perlang.ConsoleApp/Program.cs
```diff
@@ -487,6 +517,7 @@ public string[] GetSuggestions(string text, int index)
             {
                 var matchingKeywords = Scanner.ReservedKeywords
                     .Where(keyword => keyword.Key.StartsWith(text))
+                    .Where(keyword => keyword.Value != TokenType.RESERVED_WORD)
```
perlun, line 520: (Drive-by-fix: the reserved words doesn't make any sense to include in the `GetSuggestions` reuslt, since they are just "reserved for future use" in that sense.)

src/Perlang.ConsoleApp/Program.cs
perlun: Hmm... apparently not, after all (?). It doesn't look very pretty at all. We probably need to detect Windows and disable colors for now.

image

Issue #287, opened 2022-02-24
Prepare for 0.1.0 release


perlun:
1. Automate the generation of the release artifacts. This has been started, PR will hopefully come up shortly. (To be honest, I feel a bit distracted right now and it's hard to focus on spare-time programming projects. :ukraine:, you have my full support. :grimacing:)
   * One thing not yet fixed is the naming of the artifacts. Right now they get names like `perlang-0.1.0-dev.227-linux-x64.tar.gz`. It should be 0.1.0, without any suffix or anything. The same goes with the version reported by `perlang -V`.
   
     (How can we do this, i.e. distinguish between a tagged and an untagged commit?)
1. Write (or generate) some release notes. _Note_: should we keep these in the repository, somewhere under `docs`? :thinking: Perhaps `docs/release-notes/index.md`. Need to think about this part since it'll be reflect in the URI. Some examples seen on the web:
   * **Sequel**: https://github.com/jeremyevans/sequel/blob/master/CHANGELOG: The _easiest_ way to handle it do it like this: keep it in a single file, in the repo, where a new entry is just generated for each release. Harder to automate with https://github.com/softprops/action-gh-release that way, though.
   * **Ruby**: The Ruby project seems to create a news blog post for each release: https://www.ruby-lang.org/en/news/2022/02/18/ruby-3-1-1-released/
   * **Postgres JDBC**: This is a pretty nice example, where the file is readable in rendered form in GitHub (and perhaps elsewhere?): https://github.com/pgjdbc/pgjdbc/blob/master/CHANGELOG.md

   I think we'll live with not automating this part fully for now. It's not the end of the world if this part of the process is manual.

----

We have to think how it should work, though. What if... it would be something like this:

* I create a git tag each time I want to make a release, `v0.1.0` or similar. Preferably a GPG-signed tag to enhance security a bit.
* The GitHub Actions workflow detects this tag (`v*`) and creates the release. However, the release is still created as a _draft_.
* Hmm, thinking loudly as a I type here. It would surely be nice to fully automate it after all. Perhaps we'll need to go with an approach like this:
  * Create the release notes as separate files, under `docs/`, `release-notes/` or whatever.
  * Add a `make` target which concatenates them into a combined `CHANGELOG.md` or similar.
  * We can then get the best of all worlds: fully automated releases _and_ a readable full changelog in the repo. Perhaps this will be one of the better approaches.

perlun:
[...]

PR #291, opened 2022-02-27
(scripts) Handle tag `v`-prefix

perlun:
Since our SemVer tags are named things like `v0.1.0`, git describe will in these cases return a value with a `v`-prefix. This change makes the script generate valid Assembly Versions (`x.y.z`) in such scenarios, which would otherwise cause the build to fail.

Related issue: #287.

### DIFF
--- a/.github/workflows/publish-release-packages.yml
+++ b/.github/workflows/publish-release-packages.yml
@@ -34,12 +34,8 @@ jobs:
         with:
           dotnet-version: 6.0.x
 
-      - name: Get version from tag
-        id: get_version
-        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
-
-      - name: Create CommonConstants file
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-release-packages.yml
```diff
@@ -34,12 +34,8 @@ jobs:
...
           dotnet-version: 6.0.x
 
-      - name: Get version from tag
-        id: get_version
-        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
-
-      - name: Create CommonConstants file
-        run: scripts/update_common_constants.rb $(echo ${{ steps.get_version.outputs.VERSION }} | sed s/^v//)
+      - name: Re-generate auto-generated files
+        run: make auto-generated
```
perlun, line 38: Given these change in the script, we no longer need to pass in the version to the script like this. This actually means that the `CUSTOM_VERSION` support which was added in #289 is not even needed any more, but we keep it for now.

PR #297, opened 2022-03-09
(scripts) Use a correct approach for getting git commit

### LINKED ISSUE #292: GIT_COMMIT gets set incorrectly for tagged commits
After the recent [0.1.0 release](https://github.com/perlang-org/perlang/releases/tag/v0.1.0), it was discovered that `GIT_COMMIT` in `scripts/update_common_constants.rb` does not get set correctly when the latest commit is also tagged. This gives the following result in the generated binary:

```
$ "$HOME/.perlang/release/bin/perlang" -V
Perlang 0.1.0 (built from git commit v0.1.0) on .NET 6.0.2

  Number of detected (v)CPUs: 8
  Running in 64-bit mode: True
  Operating system info: Unix 5.10.0.10
```

perlun:
Fixes #292.

### DIFF
--- a/scripts/update_common_constants.rb
+++ b/scripts/update_common_constants.rb
@@ -15,7 +15,7 @@ GIT_TAG_VERSION.sub!(%r{^v}, '')
 GIT_DESCRIBE_VERSION = CUSTOM_VERSION || `git describe --tags | sed s%^dev/%% | sed s/-g.*$// | sed -E "s/-([0-9]*)$/-dev.\\1/"`.rstrip
 GIT_DESCRIBE_VERSION.sub!(%r{^v}, '')
 
-GIT_COMMIT = `git describe --always`.rstrip
+GIT_COMMIT = `git rev-parse --short HEAD`.rstrip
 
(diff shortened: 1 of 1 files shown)


scripts/update_common_constants.rb
```diff
@@ -15,7 +15,7 @@
 GIT_DESCRIBE_VERSION = CUSTOM_VERSION || `git describe --tags | sed s%^dev/%% | sed s/-g.*$// | sed -E "s/-([0-9]*)$/-dev.\\1/"`.rstrip
 GIT_DESCRIBE_VERSION.sub!(%r{^v}, '')
 
-GIT_COMMIT = `git describe --always`.rstrip
+GIT_COMMIT = `git rev-parse --short HEAD`.rstrip
```
perlun, line 18: The previous approach doesn't work correctly on tagged commits. The new approach was suggested here on SO: https://stackoverflow.com/questions/949314/how-to-retrieve-the-hash-for-the-current-commit-in-git#comment17766441_949391

PR #301, opened 2022-03-13
(solution) Upgrade SonarAnalyzer.CSharp

perlun:
This brings in support for file-scoped namespaces, supported in .NET 6: https://github.com/SonarSource/sonar-dotnet/issues/4731

Required for #300 to build successfully.


perlun:
(I'll skip the changelog entry for this one, since it's more of an "internal" change that doesn't affect the generated binaries in any way whatsoever. Including it would mostly just be "noise" in that sense.)

PR #306, opened 2022-03-20
(parser) Fix broken support for certain negative integer literals

### LINKED ISSUE #302: Certain negative integer literals are broken
I just realized that this logic is flawed:

https://github.com/perlang-org/perlang/blob/11f48ce41c7507e26d086e3c7992a9ea2be42fb2/src/Perlang.Parser/Scanner.cs#L484-L503

The problem is that this does not handle all negative numbers correctly, which is evidenced by running  the following statement in the REPL:

```
> var v = -2147483648;
[line unknown] Unsupported target type System.UInt32
```

perlun:
This commit changes the logic to do the parsing of numeric literals at parse-time instead of scan-time, where we have better access to the actual context of how the literal is being used. This makes it possible to avoid making expressions like "-100" be `Expr.UnaryPrefix` and instead be a simple `Expr.Literal` with the value of the expected type, even for the edge cases described in #302.

Fixes #302

### DIFF
--- a/src/Perlang.Parser/NumberParser.cs
+++ b/src/Perlang.Parser/NumberParser.cs
@@ -1,0 +1,141 @@
+using System;
+using System.Globalization;
+using System.Numerics;
+using Perlang.Extensions;
+
+#nullable enable
+namespace Perlang.Parser;
+
+internal static class NumberParser
+{
+    public static object Parse(NumericToken numericToken)
+    {
+        string numberCharacters = (string)numericToken.Literal!;
+
+        if (numericToken.IsFractional)
+        {
+            // TODO: This is a mess. We currently treat all floating point values as _double_, which is insane. We
+            // TODO: should probably have a "use smallest possible type" logic as below for integers, for floating point
+            // TODO: values as well. We could also consider supporting `decimal` while we're at it.
+
+            // The explicit IFormatProvider is required to ensure we use 123.45 format, regardless of host OS
+            // language/region settings. See #263 for more details.
+            return Double.Parse(numberCharacters, CultureInfo.InvariantCulture);
+        }
+        else
+        {
+            // Any potential preceding '-' character has already been taken care of at this stage => we can treat
+            // the number as an unsigned value. However, we still try to coerce it to the smallest signed or
+            // unsigned integer type in which it will fit (but never smaller than 32-bit). This coincidentally
+            // follows the same semantics as how C# does it, for simplicity.
+
+            BigInteger value = numericToken.NumberBase switch
+            {
+                NumericToken.Base.DECIMAL =>
+                    BigInteger.Parse(numberCharacters, numericToken.NumberStyles),
+
+                NumericToken.Base.BINARY =>
+                    Convert.ToUInt64(numberCharacters, 2),
+
+                NumericToken.Base.OCTAL =>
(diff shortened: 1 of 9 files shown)


src/Perlang.Parser/NumberParser.cs
```diff
@@ -0,0 +1,141 @@
...
+                // through this code path.
+                return negativeValue;
+            }
+        }
+        else
+        {
+            throw new ArgumentException($"Type {value.GetType().ToTypeKeyword()} not supported");
+        }
+    }
+}
```
perlun, line 141: Extracting all of this into a separate class has the nice side effect of making it more easy to test this in isolation in a unit test, if needed. Right now, there's no pressing need; we have integration tests which ensures this code gets exercised thoroughly. When we add support for new integer types, it could make sense to add unit tests if it makes sense. In general, I feel that integration tests have the greatest value in this kind of project anyway, so this is just a vague idea/suggestion.

src/Perlang.Parser/PerlangParser.cs
```diff
@@ -564,7 +564,18 @@ private Expr UnaryPrefix()
...
+                // of retaining it as a unary prefix expression. See #302 for some more details about why this was
+                // changed.
+                if (@operator.Type == MINUS && right is Expr.Literal rightLiteral)
+                {
+                    return new Expr.Literal(NumberParser.MakeNegative(rightLiteral.Value!));
+                }
+                else
+                {
+                    return new Expr.UnaryPrefix(@operator, right);
+                }
```
perlun, line 578: :point_up: This is an important aspect of this PR, where the logic in `NumberParser.MakeNegative` is part of solving #302 in a good way. (ensuring numbers like `-2147483648` are `System.Int32` and not `System.Int64`)

src/Perlang.Tests.Integration/Operator/Exponential.cs
```diff
@@ -7,13 +7,13 @@
...
+    ///
+    /// The type of the returned value is varies depending on the input types.
+    /// </summary>
     public class Exponential
     {
-        //
-        // Tests for the ** (exponential) operator. The operator works pretty much like in Ruby. An interesting detail
-        // about it is that the returned value is a BigInteger, regardless of the size of the input operands. This
-        // may sometimes be impractical, but for simple/REPL scenarios it is still basically useful.
-        //
```
perlun, line 16: (The `BigInteger` detail is no longer true, so this proves that "comments are lies" once again.)

src/Perlang.Tests.Integration/Typing/TypingTests.cs
```diff
@@ -102,6 +102,120 @@ public void var_declaration_in_function_detects_type_not_found()
...
+            // Int32.MinValue
+            string source = @"
+                var i = -2147483648;
+
+                print i.get_type();
+            ";
+
+            string result = EvalReturningOutputString(source);
+            Assert.Equal("System.Int32", result);
+        }
```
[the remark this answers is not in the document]
perlun: This was nice. I managed to convert _all these new tests_ into a single `[Theory]`-based-test instead. Much better and less repetitive this way, and paves the way nicely for #70. (it's really trivial to add tests for new values in the code now)

PR #307, opened 2022-03-22
(many) Add support for `uint` data type

perlun:
Related issue: #70.


src/Perlang.Parser/NumberParser.cs
```diff
@@ -60,82 +60,121 @@ public static object Parse(NumericToken numericToken)
...
+                BigInteger bigintValue => bigintValue.GetBitLength(),
+                _ => throw new ArgumentException($"Unsupported numeric type encountered: {value.GetType().Name}")
+            };
+        }
+
+        public override string? ToString()
+        {
+            return Value.ToString();
+        }
+    }
```
perlun, line 179: The changes in this file compose a significant refactoring. I think we'll extract this to a separate MR for reviewability, it's too entangled right now to be able to read this commit at a later stage and get _any_ idea of why things were done like this. :slightly_smiling_face:

src/Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -1540,17 +1540,47 @@ public VoidObject VisitWhileStmt(Stmt.While stmt)
...
+                        {
+                            // TODO: Is this safe? Do we perform enough typechecks elsewhere to feel confident enough about it?
+                            // TODO: If it's good, we should do something similar on other arm(s) as well.
+                            return (uint)(leftUint + rightInt);
+                        }
+                        else
+                        {
+                            return leftUint + rightInt;
+                        }
                     }
```
perlun, line 1584: This arm (`uint + int`) might not be relevant right now, at least not for variables:

```
> var i: uint = 1;
> var j: int = -1;
> var k = i + j;
[line 1] Error at '+': Unsupported + operand types: 'uint' and 'int'
```

Will check with integer literals next.

perlun:
Let's get this merged now, and let's hope the next data type doesn't take as much time. I've been working on this since March 13, i.e. more than 2 months. :joy: But of course, adding full test coverage for all supported type combinations (#313, #316 etc) took its time.

PR #310, opened 2022-03-25
(interpreter) Prevent assignment from double to integer types

### LINKED ISSUE #309: Assignment from `double` to `long` should generate error
At the moment, the following program doesn't generate any errors. This is wrong, and this has worked correctly previously. Let's try to figure out which commit broke it. :thinking: 

```perlang
var d: double = 8589934592.1;
var l: long = d;
```

perlun:
Fixes #309.

No changelog entry is needed on this one, since it fixes a regression introduced in #300; no release has been made where the bug was present.

### DIFF
--- a/src/Perlang.Interpreter/Typing/TypeCoercer.cs
+++ b/src/Perlang.Interpreter/Typing/TypeCoercer.cs
@@ -16,12 +16,16 @@ namespace Perlang.Interpreter.Typing
             { typeof(Int32), 32 },
             { typeof(Int64), 64 },
 
-            // Double-precision values are 64-bit but can only save 53-bit integers with exact precision
-            { typeof(Double), 32 },
-
             // In practice, even larger numbers should be possible. For the time being, I think it's quite fine if
             // bigints in Perlang are limited to 2 billion digits. :)
             { typeof(BigInteger), Int32.MaxValue }
+        }.ToImmutableDictionary();
+
+        private static ImmutableDictionary<Type, int?> FloatIntegerLengthByType => new Dictionary<Type, int?>
+        {
+            // Double-precision values are 64-bit but can only save 53-bit integers with exact precision. Since there
+            // are no "53-bit" types, we just make this the "next smaller available" bit length.
+            { typeof(Double), 32 }
         }.ToImmutableDictionary();
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Perlang.Interpreter/Typing/TypeCoercer.cs
```diff
@@ -16,14 +16,18 @@ public static class TypeCoercer
             { typeof(Int32), 32 },
             { typeof(Int64), 64 },
 
-            // Double-precision values are 64-bit but can only save 53-bit integers with exact precision
-            { typeof(Double), 32 },
-
```
perlun, line 21: This is the fix. We need to take this one out of this table, to be able to handle integer and floating-point types differently...

src/Perlang.Interpreter/Typing/TypeCoercer.cs
```diff
@@ -16,14 +16,18 @@ public static class TypeCoercer
...
             { typeof(BigInteger), Int32.MaxValue }
         }.ToImmutableDictionary();
 
+        private static ImmutableDictionary<Type, int?> FloatIntegerLengthByType => new Dictionary<Type, int?>
+        {
+            // Double-precision values are 64-bit but can only save 53-bit integers with exact precision. Since there
+            // are no "53-bit" types, we just make this the "next smaller available" bit length.
+            { typeof(Double), 32 }
+        }.ToImmutableDictionary();
+
```
perlun, line 30: ...and put it here instead.

src/Perlang.Interpreter/Typing/TypeCoercer.cs
```diff
@@ -76,7 +80,7 @@ public static bool CanBeCoercedInto(Type targetType, Type sourceType)
             // TODO: Implement more of the coercions being advertised in the XML docs. :)
 
             int? sourceSize = SignedIntegerLengthByType.TryGetObjectValue(sourceType);
-            int? targetSize = SignedIntegerLengthByType.TryGetObjectValue(targetType);
+            int? targetSize = SignedIntegerLengthByType.TryGetObjectValue(targetType) ?? FloatIntegerLengthByType.TryGetObjectValue(targetType);
```
perlun, line 83: Then, we can use it here. Note that it is **specifically** only used for the target type; conversion from `int` to `double` is valid but not the other way around.

Issue #311, opened 2022-03-29
Support division between `float` and `int`


perlun:
The following combination of operands are currently invalid:

```
> 12.0 / (9 * 60 + 1)  
[line 1] Operands must be numbers, not double and int
```

This should be valid, since `double` and `int` are perfectly able to be combined (all 32-bit positive and negative integer values can be exactly represented in `double` per the IEEE754 specification) => the type of the operation above should be `double`.

(Side note, but we should probably not postpone implementing a single-precision `float` type for too long, since it might involve changing quite a lot of the semantics if we do it too late. See #70 for the overall ticket about implementing the missing basic data types.)

perlun:
(We should check the other basic operations of math as well to ensure they all work, when this is fixed.)

perlun:
> (We should check the other basic operations of math as well to ensure they all work, when this is fixed.)

#316 made the handling of binary operators much more predictable. Now, all combinations of numeric types are *explicitly handled*. No more surprises; if something is unsupported, we have a test that verifies that it emits the expected type of error (at first, runtime errors, but as of #329 they should all be validation (compile-time) errors). Once a type combination gets supported, we always have to edit the tests => the tests will from now on **always** depict a true image of the state of what type combinations are expected to work and which ones are expected to fail.

perlun:
Fixed in #327, closing.

PR #313, opened 2022-04-10
(tests) Add full coverage for all binary operator type combinations

perlun:
This is done using a new `BinaryOperatorData` class. The purpose of this class is to have a centralized location where "all" test data for the binary operators can be placed. This way, when adding a new type you can go through all the binary operators in a single place and add test cases for the newly added type, ensuring that it works  correctly with them all.

These tests pave the way for fixing #311. The fix for these errors (and making them be `ValidationError` instead of `RuntimeError`) will be done in one or more separate PRs.


src/Perlang.Tests.Integration/Operator/Binary/Addition.cs
```diff
@@ -1,44 +1,69 @@
...
-namespace Perlang.Tests.Integration.Operator
+namespace Perlang.Tests.Integration.Operator.Binary
 {
     public class Addition
     {
-        [Fact]
-        public void addition_of_strings_performs_concatenation()
+        [Theory]
+        [MemberData(nameof(BinaryOperatorData.Plus_PlusEqual_result), MemberType = typeof(BinaryOperatorData))]
+        void addition_of_numbers_performs_numeric_addition(string i, string j, string expectedResult)
```
perlun, line 17:
```suggestion
        void with_supported_types_performs_addition(string i, string j, string expectedResult)
```

src/Perlang.Tests.Integration/Operator/Binary/Addition.cs
```diff
@@ -1,44 +1,69 @@
...
+                var i2 = {j};
+
+                print (i1 + i2).get_type();
+            ";
+
+            string result = EvalReturningOutputString(source);
+
+            result.Should()
+                .Be(expectedType);
         }
```
perlun, line 47: Should we add a `unsupported_types_emits_expected_errors`-style test here also? I think it would make sense.

perlun:
This is really a lot of work... Here's some status on the test data for the operators:

- [ ] Subtraction
- [ ] SubtractionAssignment

perlun:
> This one has quite out of hands since I started working on it some weeks (!) ago. Here's some status on the test data for the operators, mostly to help myself keep track of all the work and what parts of it are now done:

Hmm, I don't really have the energy to keep this dragging on much longer (as mentioned, it's already been going on for weeks). It's not _complete_ in that sense (not all operators have test data for all types supported by the language), but... let's live with it. :grin: Let's just merge it and fix the remaining test data in a separate PR. I'm also looking to add a new test which **enforces** that all operators have proper test data; it will make it much easier to ensure I didn't miss any type combination by mistake.

src/Perlang.Tests.Integration/Operator/Binary/Addition.cs
perlun: Has been fixed, resolving.

perlun:
> Here's some status on the test data for the operators, mostly to help myself keep track of all the work and what parts of it are now done:

> I'm also looking to add a new test which **enforces** that all operators have proper test data; it will make it much easier to ensure I didn't miss any type combination by mistake.

The proper fix to ensure that all operators have:

- Full test data for all supported type combinations _and_
- We don't forget to add test data for any operator _and_
- We keep it this way to avoid regressions in the future (especially important when #70 is being implemented). :tada: 

...is now available in #316. With the latest changes in place, it should now be CI clean. :rocket: Let's see if it in a few moments. :sweat_smile:

PR #316, opened 2022-04-20
(tests) Add `BinaryoperatorData` test

perlun:
The purpose of this test is to ensure that all type combinations are present in the parameter data in `BinaryOperatorData`.


perlun:
Note: this test is currently expected to fail, since not all type combinations are currently properly handled. This will be fixed in on or more separate PRs.

PR #327, opened 2022-05-02
(interpreter) Improved type combination support w/ binary operators

perlun:
In particular, `float`/`double`+`int`/`long` operands were often broken
(missing in the implementation since the conversion from `dynamic` to
"static code").

### DIFF
--- a/src/Perlang.Interpreter/PerlangInterpreter.cs
+++ b/src/Perlang.Interpreter/PerlangInterpreter.cs
@@ -890,6 +890,13 @@ namespace Perlang.Interpreter
             throw new RuntimeError(@operator, "Operand must be a number.");
         }
 
+        /// <summary>
+        /// Ensures that the given operands to a binary expression are numeric.
+        /// </summary>
+        /// <param name="operator">The operator for the given operand. Only used for error handling.</param>
+        /// <param name="left">The left-hand operand (evaluated value).</param>
+        /// <param name="right">The right-hand operand (evaluated value).</param>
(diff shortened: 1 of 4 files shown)


src/Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -1687,7 +1761,9 @@ public VoidObject VisitWhileStmt(Stmt.While stmt)
                 case STAR_STAR:
                     CheckNumberOperands(expr.Operator, left, right);
 
-                    if (left is float or double || right is float or double)
+                    if ((left is float or double && right is float or double) ||
+                        (left is float or double && right is int) ||
+                        (left is int && right is float or double))
                     {
                         double value = leftConvertible!.ToDouble(CultureInfo.InvariantCulture);
                         double exponent = rightConvertible!.ToDouble(CultureInfo.InvariantCulture);
```
perlun, line 1769: Unsure of why I did this change (I started working on it weeks ago but got distracted by #313 and related PRs), but I _think_ it was because `long` values cannot really be coerced to `double` like this without risk of precision loss. So that's why we wanted to narrow this down a bit, saying that `int`+`double` or `double`+`int` is fine, but `long`+`double` or vice versa is forbidden.

(In general, I think we should do a review of what type combinations we currently support with our binary operators. There are some of them being introduced now in this PR and some which were perhaps existing earlier already which might not be completely "safe" from an "avoid-precision-loss" perspective. I think I'll have to chew on that one for a while; it involves some very non-obvious tradeoffs.)

PR #329, opened 2022-05-10
(interpreter) Make operand type checks be compile-time errors

perlun:
This greatly improves the developer experience. It involves a bit of duplication, but the duplication is more or less inevitable in this case.


src/Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -904,6 +904,7 @@ private static void CheckNumberOperands(Token @operator, object? left, object? r
                 return;
             }
 
+            // TODO: Why does this still happen?
             throw new RuntimeError(@operator, $"Operands must be numbers, not {StringifyType(left)} and {StringifyType(right)}");
```
perlun, line 908: There are still some tests which make assertions on this (runtime) error, which puzzles me a bit. I'll look into this a bit further before getting this merged.

perlun: Ah, I get it now... This is a bit hairy. :see_no_evil: Here's an example test that will trigger this:

```perlang
fun greet(name: string, age: int): void {
  print "Hello " + name + ". Your age is " + age;
}

// Expected warning: [line 6] Warning at 'name': Null parameter detected
greet(null, 42);
```

The above is understood as `string` + `string`, but `string` + `null` will cause the exception above. This is not ideal; I filed a new bug about it now: #330.

PR #337, opened 2022-06-11
(many) Add support for `ulong` data type

perlun:


src/Perlang.Tests.Integration/Operator/Binary/BinaryOperatorData.cs
```diff
@@ -334,39 +334,39 @@ public static class BinaryOperatorData
...
+            new object[] { "4294967295", "9223372036854775807", "-9223372032559808512", typeof(long) }, // Negative integer overflow
+            new object[] { "4294967295", "18446744073709551615", "4294967296", typeof(ulong) }, // Positive integer overflow
             new object[] { "4294967295", "18446744073709551616", "-18446744069414584321", typeof(BigInteger) },
             new object[] { "4294967295", "12.0", "4294967283", typeof(double) },
             new object[] { "9223372036854775807", "2", "9223372036854775805", typeof(long) },
             new object[] { "9223372036854775807", "4294967295", "9223372032559808512", typeof(long) },
             new object[] { "9223372036854775807", "9223372036854775807", "0", typeof(long) },
             new object[] { "9223372036854775807", "18446744073709551616", "-9223372036854775809", typeof(BigInteger) },
-            new object[] { "9223372036854775807", "12.0", "9.223372036854776E+18", typeof(double) }, // TODO: We should make this unsupported. As can be seen in the exponential expression, the operation loses precision.
+            new object[] { "9223372036854775807", "12.0", "9.223372036854776E+18", typeof(double) },
```
perlun, line 345: I removed these "we should make this unsupported" comments note. `long + double` and `ulong + double` does indeed have precision loss, but such operations are supported in Java and C# so we can safely presume that there can be good motivations for supporting it. One way to think of it: the user who is _using_ double-precision FP math is expected to be aware of its limitations.

We don't support it for `BigInteger + double` at the moment, since it *definitely* loses precision (the fractional part would be truncated).

Issue #182, opened 2021-04-11
(consoleapp) Remember command history in REPL


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/182. The GitHub page you are currently reading will not contain the latest information on this issue.

---

GNU readline has a nice feature: it remembers the command line history _between invocations_ of a given program. This can be really useful sometimes, for example when typing a particular program in the REPL which exhibits a particular bug, doing a change, recompiling & rerunning the REPL again to see how it changes the program in question => up-arrow would be super-nice.

We don't use GNU readline at the moment, but maybe we should? How hard would it be to integrate it with Perlang as it stands? I'm not at all convinced that the "managed only" approach that most Java and C# programs take is the way to go, and we really want Perlang (programs) to be able to easily interoperate with random C libraries in fact. Maybe this would be a good candidate to trying out how hard it would be.

### Caveat

GNU readline is GPL-licensed. We must be _extremely careful_ with using GPL-licensed software as part of our tooling, since it might pull our whole codebase into it because of how the GPL works. The Perlang REPL is technically isolated from the actual _interpreter_ in Perlang, but I still think that something like BSD's `libedit` would be a safer choice if we want to avoid "the GPL trap" (sorry for this wording, I'm tired and it's late in the evening. I love the accomplishments of the FSF and their emphasis on freedom etc; I just don't vibe with the GPL in this regard.)

perlun:
### Alternatives to GNU readline

* ReadLine: https://github.com/tonerdo/readline 
    * C# GNU-Readline like library
    * MIT-licensed
    * Unmaintained, no commits pushed since June 2018
    * Works on both Unix-like systems and Windows.
* BSD libedit: https://www.thrysoee.dk/editline/, git repo exists at https://salsa.debian.org/debian/libedit
    * C library, requires P/Invoke
    * BSD-licensed (3 clause)
    * Probably only works on Unix-like systems (=not Windows)
* Linenoise: https://github.com/antirez/linenoise
    * C librarym requires P/Invoke
    * BSD-licensed (2-clause)
    * Only works on VT100-compatible systems, i.e. Unix.
    * Reputable author; @antirez is the original author of [Redis](https://redis.io/)
 
### Another option

Given the above, I'm considering that we should author a library of our own to handle this, meeting the following requirements:

[...]

PR #354, opened 2022-12-30
(consoleapp) Switch from ReadLine to getline.cs

perlun:
Fixes part of #182. :tada: While not being perfect in any way, the achievement made here in a single evening is a noticeable improvement from the previous behavior. I'm willing to merge this as a first-step.

(Couldn't have been done without the previous effort of @migueldeicaza in https://github.com/mono/mono/blob/master/mcs/tools/csharp/getline.cs, thanks a lot. :pray:)


perlun:
Missing before merge: add changelog entry.

Issue #270, opened 2022-02-06
Support array/dictionary indexing


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/270. The GitHub page you are currently reading will not contain the latest information on this issue.

---

#269 adds some new methods, which shows the need for some new language elements:

```perlang
var env = Libc.environ();
print env.get_Item("PATH");
```

The above _works_, but is anywhere near the elegance we are striving for. Here's the suggested syntax that should be supported:

```perlang
var env = Libc.environ();
print env["PATH"];
```

...in other words, the same as in C#, Ruby, Python and JavaScript (but interestingly enough, **not** in Java; or rather, Java only supports it for built-in arrays (`String[]`, `Object[]` etc), not for other types like `Map<TKey, TValue>`.)

Initially, we can make it work just for `Dictionary<TKey, TValue>`, but in the longer run it would be great to support other indexable types as well (`List<T>` and arrays - the latter would need language support to be able to define arrays though, and the former cannot be instantiated from Perlang since we currently don't support neither the `new` keyword nor using generics. :see_no_evil: However, as noted above, we might still _run into_ arrays or lists being returned from library-level functions, so perhaps we should support such indexing in the language right away. :thinking:).

perlun:
This feels like one of the more critical things we are missing at the moment. Let's aim for including it in 0.3.0; we can always bump it a few more versions if needed.

perlun:
Retained open despite #363 being merged because of the missing features: https://github.com/perlang-org/perlang/pull/363/files#r1104893169

perlun:
What we have now is good enough for getting 0.3.0 out the door. Moving the remaining work to the 0.4.0 iteration.

perlun:
This is not yet working for compiled mode (#406). We should fix that, hopefully in the next 0.6.0 iteration.

PR #363, opened 2023-01-23
(many) Add support for `string` and `IDictionary` indexing

perlun:
Related to #270.

### DIFF
--- a/src/Perlang.Interpreter/Typing/TypeResolver.cs
+++ b/src/Perlang.Interpreter/Typing/TypeResolver.cs
@@ -1,5 +1,6 @@
 #nullable enable
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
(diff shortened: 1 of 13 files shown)


src/Perlang.Interpreter/Typing/TypeResolver.cs
```diff
@@ -388,6 +389,75 @@ public override VoidObject VisitCallExpr(Expr.Call expr)
...
+                            expr.ClosingBracket,
+                            $"'string' cannot be indexed by '{argumentType.ToTypeKeyword()}'")
+                        );
+                    }
+
+                    expr.TypeReference.ClrType = typeof(char);
+                    break;
+
+                // TODO: arrays
+                // TODO: support lists
```
perlun, line 425: This will be left unsupported for now. We don't have a way to define user-defined arrays or lists in the language, and I'm not aware of any `stdlib` method that currently needs this. :thinking: We can easily add it when we see the need for it, and I don't want to add support for it without zealously adding integration tests for it at the same time... :slightly_smiling_face: 

Will keep #270 open precisely because of this.

Issue #370, opened 2023-02-23
Implement `string` interface with smart handling of string literals


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/370. The GitHub page you are currently reading will not contain the latest information on this issue.

---

### General proposal

The `string` type should automatically handle different types of underlying string representations

### Current state

Currently, code like this:

`var s: string = "foo"`

...will declare the `s` variable as a standard, .NET `string`. The underlying binary representation is UTF-16 (LE, on LE systems like `amd64`).

### Suggested changes

- Add an `IString` interface:

```c#
interface IString
{
    bool is_utf8();
    bool is_ascii();
}
```

- Implement three different types of strings: `AsciiString`, `Utf8String` and `Utf16String`:

```c#
class AsciiString : IString, IIndexable<AsciiChar>
{
    bool is_utf8() { return true; }   // All ASCII strings are also valid UTF-8 strings
    bool is_ascii() { return true; }

    // Returns the backing byte array. We might need both a copying and non-copying form for performance.
    byte[] as_bytes() { ... }

    AsciiString as_ascii() { return this; }

    // Can be implemented very efficiently; either by copying the backing byte array or by "taking
    // over ownership" of it (à la Rust). Either way, the data does not need to be re-parsed in any
    // way.
    Utf8String as_utf8() { ... }

    // Can also be implemented fairly efficiently, since this is just expanding each character to
    // two bytes.
    Utf16String as_utf16() { ... }
}

class Utf8String : IString
{
    bool is_utf8() { return true; }

    // O(n) operation to determine whether the string is ASCII clean or not. Can be done on first 
    // call or at construction time, but doing a tradeoff depending on the string length might also 
    // make sense. Probably best to start with memoized implementation on first call and iterate
    // from that.
    bool is_ascii() { ... }

    // Will check if is_ascii() is true; if yes, can convert to ASCII by just copying the backing
    // byte array. If no, will throw exception.
    bool as_ascii() { ... }

    Utf8String as_utf8() { return this; }

    // O(n) operation. Will check if is_ascii() is true; if yes, can use similar semantics as
    // AsciiString.as_utf16(). If no, will need to parse each UTF-8 character and transform it
    // to the appropriate UTF16-LE representation.
    Utf16String as_utf16() { return this; }
}

// Note: uses UTF-16LE under the hood.
class Utf16String : IString, IIndexable<Utf16Char>
{
[...]

Issue #376, opened 2023-03-12
Rewrite the `Base64` class to not depend on the .NET BCL


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/376. The GitHub page you are currently reading will not contain the latest information on this issue.

---

At the moment, our Base64 encoding (and decoding) heavily relies on the .NET BCL:

https://github.com/perlang-org/perlang/blob/a342fbfae839c70fb3ca9a34b4e1e854be49fa9f/src/Perlang.Stdlib/Stdlib/Base64.cs#L22-L27

While this is an easy way to get started with adding functionality to the Perlang stdlib, it's less than perfect. Especially now that we are working towards #370, it'll actually become less convenient since it will involve conversion from Perlang strings to .NET strings (a form of "marshalling" if you will).

Investigate how hard it would be to do base64 operations ourselves. The ideal would be to use an optimized C/C++ implementation and call into that; if we start investigating this, we should make some simple benchmarks to see how the different implementations perform compared to each other.

This aligns well with the overarching goal of not making Perlang be overly dependent on the .NET platform itself; ideally, I don't want to close any doors right now as for the future of "compiled" Perlang. It may compile to MSIL, it may also compile to ELF/Win32 PE binaries. Both of these should remain possible. The recent steps in the first #370 implementation PR (#377) where we move towards using unmanaged `byte*` pointers for the string data seems well-aligned with this goal.

### Potential implementations

(courtesy of https://stackoverflow.com/questions/342409/how-do-i-base64-encode-decode-in-c)

- [libb64](https://libb64.sourceforge.net/)
- OpenSSL
- glib

PR #377, opened 2023-03-12
(many) Implement `AsciiString` for ASCII-only strings

perlun:
Related to #370, this fixes the first parts of this issue.

### DIFF
--- a/release-notes/v0.4.0.md
+++ b/release-notes/v0.4.0.md
@@ -4,6 +4,9 @@
 ### Added
 
 ### Changed
+#### Data types
+- Implement `AsciiString` for ASCII-only strings [[#377][377]]
+
 #### Maintenance
 - Bump TngTech.ArchUnitNET.xUnit from 0.5.0 to 0.10.5 [[#366][366]]
 - Bump xunit from 2.4.1 to 2.4.2 [[#369][369]]
@@ -15,4 +18,5 @@
 [366]: https://github.com/perlang-org/perlang/pull/366
 [369]: https://github.com/perlang-org/perlang/pull/369
 [374]: https://github.com/perlang-org/perlang/pull/374
+[377]: https://github.com/perlang-org/perlang/pull/377
 [380]: https://github.com/perlang-org/perlang/pull/380
--- a/src/Perlang.Stdlib/Internal/Extensions/TypeExtensions.cs
+++ b/src/Perlang.Stdlib/Internal/Extensions/TypeExtensions.cs
@@ -1,7 +1,8 @@
 using System;
 using System.Numerics;
+using Perlang.Lang;
 
-namespace Perlang.Extensions
+namespace Perlang.Internal.Extensions
 {
     /// <summary>
     /// Extension methods for <see cref="Type"/>.
@@ -20,7 +21,8 @@ namespace Perlang.Extensions
                 { } when type == typeof(Single) => "float",
                 { } when type == typeof(Double) => "double",
                 { } when type == typeof(NullObject) => "null",
-                { } when type == typeof(String) => "string",
+                { } when type == typeof(Lang.String) => "string",
+                { } when type == typeof(AsciiString) => "AsciiString",
 
                 // TODO: add bool here
                 { } when type.IsAssignableTo(typeof(IPerlangFunction)) => "function",
--- a/src/Perlang.Stdlib/Lang/AsciiString.cs
+++ b/src/Perlang.Stdlib/Lang/AsciiString.cs
@@ -1,0 +1,271 @@
+#nullable enable
+#pragma warning disable S112
+#pragma warning disable SA1300
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using Perlang.Internal;
+using Perlang.Stdlib;
+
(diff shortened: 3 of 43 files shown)


perlun:
We now have an implementation which mostly works (at least on POSIX platforms; I think the `toupper` P/Invoke needs a minor adjustment to work on Windows as well). However, the following issues remain to be settled before this is even close to being mergeable:

* We leak memory like crazy. Each `AsciiString` created leaks memory which is never reclaimed (until the process exits). We should add our own `Allocator` class which takes care of this, and frees memory in a controlled way on process shutdown. Subsequent improvements will be to implement the Perlang memory model, which is yet to be defined.

* `AsciiString` is currently defined in `Perlang.Common`, whereas it should rightfully be in `Perlang.Stdlib`. However, moving it there will cause some interesting dependency-related issues that remain to be resolved.

release-notes/v0.4.0.md
```diff
@@ -5,4 +5,11 @@
...
 ### Changed
 
+#### Maintenance
+- Bump TngTech.ArchUnitNET.xUnit from 0.5.0 to 0.10.5 [[#366][366]]
+- Bump xunit from 2.4.1 to 2.4.2 [[#369][369]]
+
 ### Fixed
+
+[366]: https://github.com/perlang-org/perlang/pull/366
+[369]: https://github.com/perlang-org/perlang/pull/369
```
perlun, line 15: TODO: Move these to a separate commit, should not be included in this PR.

src/Perlang.Stdlib/Internal/Extensions/TypeExtensions.cs
```diff
@@ -1,7 +1,8 @@
 using System;
 using System.Numerics;
+using Perlang.Lang;
 
-namespace Perlang.Extensions
+namespace Perlang.Internal.Extensions
```
perlun, line 5: As part of fixing this problem:

> `AsciiString` is currently defined in `Perlang.Common`, whereas it should rightfully be in `Perlang.Stdlib`. However, moving it there will cause some interesting dependency-related issues that remain to be resolved.

...we moved certain classes (like this one) out of `Perlang.Common` and into an "internal" namespace in `Perlang.Stdlib`. They really don't belong in the (facing) stdlib at all; in the future, I envision moving these to some `Perlang.CommonStdlib` assembly or similar. ("shared code between both Perlang parser and interpreter which should not be publicly exposed to user code")

src/Perlang.Stdlib/Lang/AsciiString.cs
```diff
@@ -0,0 +1,228 @@
...
+    // TODO: of .NET strings.
+
+    private unsafe AsciiString(IReadOnlyList<char> chars)
+    {
+        // TODO: Use jemalloc instead. Requires creating a NuGet package of it for easiest consumption from this project.
+        // TODO: See https://github.com/dotnet/runtime/issues/11404 and https://github.com/allisterb/jemalloc.NET for
+        // TODO: some prior art.
+        //
+        // TODO: Use a wrapper for keeping track of our allocations, and free all unmanaged memory on VM exit. Something
+        // TODO: like AppDomain.CurrentDomain.DomainUnload could probably do.
```
perlun, line 90: TODO: This one needs to be taken care of _before_ this PR is merged.

src/Perlang.Stdlib/Lang/AsciiString.cs
```diff
@@ -0,0 +1,228 @@
...
+
+        return hashCode;
+    }
+
+    // TODO: Replace the need for this by making our "print" keyword not use
+    // TODO: ToString() for this class. Ideally, we could just call "puts" with
+    // TODO: a pointer to the `bytes` array straight away; for ASCII strings
+    // TODO: this would work on both Windows and POSIX systems. For UTF-8
+    // TODO: strings, it would *not* work on Windows though so we would need to
+    // TODO: special-case that platform... :|
```
perlun, line 176: (In a subsequent PR)

src/Perlang.Stdlib/Stdlib/Base64.cs
```diff
@@ -33,19 +34,23 @@ public static string Encode(string plainText)
...
+            string base64String = base64Data.ToString();
 
             if (numPaddingCharacters != 0)
             {
-                base64Data += new string('=', 4 - numPaddingCharacters);
+                base64String += new string('=', (int)(4 - numPaddingCharacters));
             }
 
-            var data = Convert.FromBase64String(base64Data);
+            var data = Convert.FromBase64String(base64String);
```
perlun, line 53: These fixes are an inefficient interim solution until we investigate #376.

release-notes/v0.4.0.md
perlun: Fixed locally, resolving.

src/Perlang.Stdlib/Lang/AsciiString.cs
perlun: Fixed & tested locally. We now have some basic logic in place which "garbage collects" at the end of the process. This is clearly unsuitable for many "real-world" applications, but we'll defer fixing this properly until #378 is taken care of. Considering "done" for now.

PR #389, opened 2023-04-21
(many) Use Perlang strings and native C functions for writing to stdout

perlun:
See #390 for the over-arching goal.

### DIFF
--- a/src/Perlang.ConsoleApp/PerlangConsole.cs
+++ b/src/Perlang.ConsoleApp/PerlangConsole.cs
@@ -1,0 +1,117 @@
+#nullable enable
+using System;
+using System.CommandLine.IO;
+using System.Runtime.InteropServices;
+using System.Text;
+using Perlang.Exceptions;
+using Perlang.Stdlib;
+using String = Perlang.Lang.String;
+
+namespace Perlang.ConsoleApp;
+
+// Shamelessly inspired by System.CommandLine.IO.SystemConsole,
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+public class PerlangConsole : IPerlangConsole
+{
+    public PerlangConsole()
+    {
+        Error = StandardErrorStreamWriter.Instance;
+        Out = StandardOutStreamWriter.Instance;
+    }
+
+    /// <inheritdoc />
+    public IStandardStreamWriter Error { get; }
+
+    /// <inheritdoc />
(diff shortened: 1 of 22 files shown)


src/Perlang.Interpreter/Internals/AstPrinter.cs
```diff
@@ -19,12 +19,14 @@ namespace Perlang.Interpreter.Internals
     /// </summary>
     internal class AstPrinter : Expr.IVisitor<string>, Stmt.IVisitor<string>
     {
+        // TODO: Return Perlang string instead
```
perlun, line 22: #390

src/Perlang.Parser/PerlangParser.cs
```diff
@@ -663,25 +662,10 @@ private Expr Primary()
...
-                        // TODO: https://github.com/perlang-org/perlang/issues/370
-                        return new Expr.Literal(Utf8String.from(s));
-                    }
-                }
-
-                // All characters in the string are ASCII safe => represent this
-                // string as an AsciiString in the created Literal.
-                return new Expr.Literal(AsciiString.from(chars));
+                Lang.String nativeString = Lang.String.from(s);
+                return new Expr.Literal(nativeString);
```
perlun, line 668: The code that used to be here was pretty much inlined in `String.from`, to make it be generically usable from elsewhere.

src/Perlang.ConsoleApp/PerlangConsole.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// <inheritdoc />
+    // TODO: Can we do this without relying on System.Console?
+    public bool IsOutputRedirected => Console.IsOutputRedirected;
+
+    /// <inheritdoc />
+    public bool IsInputRedirected => Console.IsInputRedirected;
+
+    public void WriteStdoutLine(String s)
+    {
+        int bytesWritten = Libc.Internal.write(Libc.Internal.STDOUT_FILENO, s.Bytes, (int)s.Length);
```
perlun, line 41: Interestingly enough, this approach seems to work on both Linux and windows (tested on Windows 10). :astonished: If we don't need to write platform-dependent code, we _should_ obviously not be writing platform-dependent code...

(Reading keypresses might be a bigger challenge. We'll get to that eventually, trying to do without `Console.ReadKey`)

Issue #406, opened 2023-08-12
Implement compiled mode for Perlang


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/406. The GitHub page you are currently reading will not contain the latest information on this issue.

---

In https://github.com/perlang-org/perlang/discussions/396, I described the recent events leading up to me trying out what LLVM can do for us, in terms of _making it possible to run Perlang programs completely independent of the .NET platform_.

After that comment was written, and some discussions I had with an old friend of mine (@diwic - thanks a lot to you! :pray:), I started hacking on this and doing a little experiment: _How hard would it be to write a compiler for Perlang, which emits C++ code, compiles this code, and then runs the end result_? This is obviously not the "final solution" in any way and it is admittedly a bit clumsy. Still, if it was [good enough for Bjarne Strousrup](https://www.stroustrup.com/bs_faq.html#bootstrapping), it ought to be good enough for me as well. (Naturally, Strousrup's preprocessor and later Cfront compiler didn't emit _C++_, but you get the picture.)

I'm setting the milestone for this to 0.4.0, but naturally, given the sheer size of this task, the compiler will in no way be _complete_ in 0.4.0. But it'll probably work to the point where I feel comfortable about pushing it out to the public.

### Rough steps
- [x] Implement a compiler which translates the syntax tree for all/most valid Perlang programs into valid C++ code, and compiles and runs the result: https://github.com/perlang-org/perlang/pull/409
- [x] Implement a C/C++-based stdlib to support the above: #407.
  - Make it possible to write unit and/or integration tests for this. We probably have to write these in C or C++ for now. [cmocka](https://cmocka.org/) is a useful unit test library for C that I have used elsewhere.
  - [x]  Add support for `BigInt`: #415
  - [x] Distribute the (compiled) `stdlib` along with snapshot builds. This involves a bit of complexity, since native C++ code has historically only been able to compile on the same platform as the CI job is running on. We'll need to investigate if `clang` makes this easier for us.
    - In line with the next point, I think it's fine if we are Linux and `amd64`-only at this point (in compiled mode). In other words, we'll provide a Linux `amd64` binary of the `stdlib` for now and emit an error message on other platforms stating that experimental compilation is not yet supported.
[...]

Issue #398, opened 2023-05-31
Investigate Perlang-installer failing on Windows


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/398. The GitHub page you are currently reading will not contain the latest information on this issue.

---

As of #389 being merged, we have started seeing a very peculiar error when installing on Windows: https://github.com/perlang-org/perlang/actions/runs/5094377058/jobs/9242418443

I have looked at the code and there's nothing obviously wrong with it: https://github.com/perlang-org/perlang/blob/1f05e2a3219aa539f391641ba046c27d376e177f/src/Perlang.Stdlib/Stdlib/Libc.cs#L61-L66

This is exactly what it should be doing. Still, with the above we get failures like this in CI:

    Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
     ---> System.DllNotFoundException: Unable to load DLL 'libc' or one of its dependencies: The specified module could not be found. (0x8007007E)
       at Perlang.Stdlib.Libc.Internal.write(Int32 fd, IntPtr s, Int32 count) in /home/runner/work/perlang/perlang/src/Perlang.Stdlib/Microsoft.Interop.LibraryImportGenerator/Microsoft.Interop.LibraryImportGenerator/LibraryImports.g.cs:line 27
       at Perlang.ConsoleApp.PerlangConsole.StandardOutStreamWriter.Write(String value) in /home/runner/work/perlang/perlang/src/Perlang.ConsoleApp/PerlangConsole.cs:line 75
       at Perlang.ConsoleApp.Program.<>c__DisplayClass11_0.<MainWithCustomConsole>b__4(ParseResult parseResult, IConsole _) in /home/runner/work/perlang/perlang/src/Perlang.ConsoleApp/Program.cs:line 175
       at System.RuntimeMethodHandle.InvokeMethod(Object, Void**, Signature, Boolean)
       at System.Reflection.MethodInvoker.Invoke(Object, IntPtr*, BindingFlags)

### Thoughts

I'm wondering if this is a new use case we haven't seen previously. :thinking: Until now, we have been _executing_ Windows-only code on Windows, which means that tests dependent on functionality different on Windows (examples: #258 and #269) has been working.

Now, we are talking about something slightly different though: we are cross-compiling for Windows _from Linux_. I'm not sure the `_WINDOWS` define is set in this case, which could very likely be part of the problem here.

[...]

PR #407, opened 2023-09-01
(stdlib) Add C++-based stdlib project

perlun:
This is part of the overall #406 issue (`Implement compiled mode for Perlang`). Not ready for merge, but we're getting there, slowly but steadily...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,5 +24,10 @@ jobs:
       - name: Re-generate auto-generated files
         run: make auto-generated
 
+      # Needed to run tests, when they are running in compilation mode
+      # (PERLANG_EXPERIMENTAL_COMPILATION=true)
+      - name: Rebuild stdlib
+        run: make stdlib
+
       - name: Run test suite
         run: dotnet test --configuration Release --verbosity normal
--- a/README.md
+++ b/README.md
@@ -146,6 +146,16 @@ editor](https://github.com/mono/mono/blob/main/mcs/tools/csharp/getline.cs),
 dual-licensed under the terms of the MIT X11 license or the Apache License 2.0.
 Copyright (c) 2008 Novell, Inc. Copyright (c) 2016 Xamarin Inc.
 
+[src/stdlib/src/double-conversion](src/stdlib/src/double-conversion) includes
+content from the Google `double-conversion` library, available at
+https://github.com/google/double-conversion, licensed under the BSD 3-Clause
+"New" or "Revised" License. Copyright 2006-2011, the V8 project authors.
+
+[src/stdlib/src/fmt](src/stdlib/src/fmt) includes content from the `{fmt}`
+library, available at https://github.com/fmtlib/fmt, licensed under the MIT
+license. Copyright (c) 2012 - present, Victor Zverovich and `{fmt}`
+contributors.
+
 ## Disclaimer
 
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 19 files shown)


perlun:
(The CI failure is because of https://github.com/perlang-org/perlang/issues/398; we might have to disable the `install_windows` job in CI for now.)

README.md
```diff
@@ -146,6 +146,11 @@ editor](https://github.com/mono/mono/blob/main/mcs/tools/csharp/getline.cs),
 dual-licensed under the terms of the MIT X11 license or the Apache License 2.0.
 Copyright (c) 2008 Novell, Inc. Copyright (c) 2016 Xamarin Inc.
 
+[src/stdlib/src/double-conversion](src/stdlib/src/double-conversion) includes
+content from the Google `double-conversion` library, available at
+https://github.com/google/double-conversion, licensed under the BSD 3-Clause
+"New" or "Revised" License. Copyright 2006-2011, the V8 project authors.
+
```
perlun, line 153: We'll likely be able to drop this, in favor of the fmt lib: https://github.com/fmtlib/fmt. A few issues remains to be settled though, to make it "compatible enough" with `Double.ToString()` in .NET.

src/stdlib/src/print.cpp
```diff
@@ -0,0 +1,62 @@
...
+    void print(float f)
+    {
+        fmt::println("{}", f);
+    }
+
+    void print(double d)
+    {
+        fmt::println("{}", d);
+    }
+}
```
perlun, line 62: :point_up: This is one of the more important parts in this PR. It features the implementation of the `perlang::print()` methods, with various overloads for the different primitive types we support.

src/stdlib/test/double-conversion/test-fast-dtoa.cc
```diff
@@ -0,0 +1,395 @@
...
+            CHECK_EQ(current_test.representation, buffer.start());
+        }
+        // The precomputed numbers contain many entries with many requested
+        // digits. These have a high failure rate and we therefore expect a lower
+        // success rate than for the shortest representation.
+        CHECK(succeeded*1.0/total > 0.85);
+        // However with less than 15 digits almost the algorithm should almost always
+        // succeed.
+        CHECK(succeeded_15*1.0/total_15 > 0.9999);
+}
```
perlun, line 395: A lot of the `double-conversion` code, if not all, should be able to be dropped. We ended up not using it in production (going with https://github.com/fmtlib/fmt instead). The tests are still present though, so we'll need to revisit them a bit.

perlun: This has been mostly done. I'll resolve this now and write a followup comment on one of the remaining parts.

.github/workflows/test.yml
```diff
@@ -24,5 +24,10 @@ jobs:
       - name: Re-generate auto-generated files
         run: make auto-generated
 
+      # Needed to run tests, when they are running in compilation mode
+      # (PERLANG_EXPERIMENTAL_COMPILATION=true)
+      - name: Rebuild stdlib
+        run: make stdlib
+
```
perlun, line 31: This runs the _compilation_ of the stdlib. We could also run the tests, but they are not fully working yet (and even when they are, there will be some more work to get them working on macOS). Let's see if we merge this now already or if we fix this first.

src/stdlib/test/double-conversion/cctest.h
```diff
@@ -0,0 +1,160 @@
...
+  TestFunction* callback_;
+  const char* file_;
+  const char* name_;
+  const char* dependency_;
+  bool enabled_;
+  static CcTest* last_;
+  CcTest* prev_;
+};
+
+#endif  // ifndef CCTEST_H_
```
perlun, line 160: For now, let's be a bit lazy and use the `cctest` test helper from https://github.com/google/double-conversion. In the long run, we should switch to https://github.com/catchorg/Catch2.

PR #409, opened 2023-09-29
(compiler) Add first steps towards experimental compiler

perlun:
- [x] All tests are now passing in compiled (and interpreted) mode.
- [x] CI settings have been adjusted so that we test interpreted mode on all supported platforms, and in addition to this, we test compiled mode on Ubuntu

What remains to be done:

- [ ] Compile and include the `stdlib` along with our release distributions, so that compiled mode is usable in these as well. This will be done in a followup MR.

PR #413, opened 2023-10-22
(interpreter) Make `int == double` use value-type semantics

perlun:
This is an important preparation for https://github.com/perlang-org/perlang/pull/409, where we will use the same semantics. Doing the change already not for interpreted mode makes things easier, since we can then use the same tests for both interpreted and compiled mode.

### DIFF
--- a/src/Perlang.Interpreter/PerlangInterpreter.cs
+++ b/src/Perlang.Interpreter/PerlangInterpreter.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Numerics;
 using System.Reflection;
 using System.Text;
+using Microsoft.CSharp.RuntimeBinder;
 using Perlang.Attributes;
 using Perlang.Exceptions;
 using Perlang.Internal.Extensions;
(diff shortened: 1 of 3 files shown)


src/Perlang.Interpreter/PerlangInterpreter.cs
```diff
@@ -948,10 +949,23 @@ private static bool IsEqual(object? a, object? b)
...
             if (a == null)
             {
+                // null != non-null
                 return false;
             }
 
-            return a.Equals(b);
+            // Note: must use dynamic here since `object == object` behaves differently than e.g. `int == double`,
+            // tragically enough. More details:
+            // https://learn.microsoft.com/en-us/dotnet/api/System.Object.Equals?view=net-7.0#notes-for-callers
```
perlun, line 958: Here's a clickable version of that link: https://learn.microsoft.com/en-us/dotnet/api/System.Object.Equals?view=net-7.0#notes-for-callers

What's particularly unintuitive in C# is perhaps the difference between these:

```c#
int i = 12;
int j = 12;

object i2 = i;
object j2 = j;

bool b = i == j;   // true
bool c = i2 == j2; // false
```

I believe this is because of [boxing](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/types/boxing-and-unboxing), which is necessary to be able to store a value type in an `object` variable, but it's still hard to argue that the above provides an _extremely_ confusing developer experience. The interesting question is: would it have been reasonably possible to provide any alternative semantics that would have worked _better_ in this regard, without any other adverse effects? :thinking:

The thing is that C# is fairly advanced in this case; much more similar to C++ than e.g. Java, with support for overloading the `==` operator and things like this. That is a very nice thing for certain use cases, but I guess it also comes at a price.

PR #414, opened 2023-10-30
(stdlib) Use better format in `float` and `double` conversion to string

perlun:
The main benefit of this new format is that it's "more" predictable, and we can more easily mimic the exact byte-by-byte output in the C++ implementation.

The downside is that we *may* have problems with roundtripping to string and back with this format, because of precision loss. Let's live with this for now; it's (for now) more important that all integration tests pass in compiled and interpreted mode. .NET Framework seems to have been using G7 and G15 by default, so it can't be that bad. ;)


perlun:
I'm marking this a Draft for now; we want to ensure we can really produce the exact same format in #409 (on top of this MR) before we consider it worth merging.

release-notes/v0.4.0.md
```diff
@@ -12,6 +12,7 @@
 #### Data types
 - Implement `AsciiString` for ASCII-only strings [[#377][377]]
 - Implement `Utf8String` for strings with non-ASCII content [[#385][385]]
+- Use more predictable format in `float` and `double` conversion to string [[#414][414]]
```
perlun, line 15: TODO: Settle on a title for this change and use it consistently, both here and in the commit message. :innocent:

perlun: Fixed, resolving. I realized now why this was inconsistent: the commit message was too long, causing GitHub to wrap it. (Should probably add a pre-commit check locally to prevent me from pushing commits which GitHub will wrap or something...)

perlun:
> I'm marking this a Draft for now; we want to ensure we can really produce the exact same format in #409 (on top of this MR) before we consider it worth merging.

This has now been verified. I'm scheduling this for merge in a moment. :tada:

PR #422, opened 2024-01-13
(compiler) Support `**` for `BigInt` values

perlun:


src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
```diff
@@ -848,14 +848,6 @@ private void AddGlobalClass(string name, PerlangClass perlangClass)
...
                     currentMethod.Append($"{leftCast}pow({expr.Left.Accept(this)}, {rightCast}{expr.Right.Accept(this)})");
                 }
-                else if (expr.Left.TypeReference.ClrType == typeof(BigInteger) ||
-                    expr.Right.TypeReference.ClrType == typeof(BigInteger))
-                {
-                    // TODO: We need something lke BigInteger.Pow() in .NET to be able to accomplish this. Perhaps do
-                    // TODO: like the .NET folks and implement something using this approach?
-                    // TODO: https://en.wikipedia.org/wiki/Exponentiation_by_squaring
-                    throw new NotImplementedInCompiledModeException("** operator for BigInteger is not yet supported in compiled mode");
-                }
```
perlun, line 858: Don't know what I was thinking when I added this. :see_no_evil: `BigInt ** BigInt` is not even supported in interpreted mode..

src/stdlib/CMakeLists.txt
```diff
@@ -22,6 +22,14 @@ target_include_directories(
...
 )
 
+target_compile_options(
+        stdlib PRIVATE -Wall -Wextra -Werror
+
+        # TODO: Remove this. We have some current issues with %lld (long long int vs int64_t) that needs to be
+        # TODO: resolved for this. The problem is IIRC that Linux and macOS treated this differently.
+        -Wno-error=format
+)
+
```
perlun, line 32: Deferring the `-Werror=format` issue for later. For now, enabling `-Wall` and `-Werror` is the important part, since without it you can do crazy things like _forgetting to return values_ from a non-`void` function.. and the compiler will happily try to compile your program anyway. :woozy_face:

src/stdlib/src/stdlib.hpp
```diff
@@ -25,4 +25,22 @@ namespace perlang
     void print(const BigInt& bigint);
     void print(float f);
     void print(double d);
+
+    BigInt BigInt_pow(const BigInt& value, int exponent)
+    {
+        // "With constant auxiliary memory" implementation as given in
+        // https://en.wikipedia.org/wiki/Exponentiation_by_squaring
```
perlun, line 32: This is actually not true, we currently have the plain recursive implementation. I'll fix this now to use the slightly better version described on that page.

perlun: Fixed and amended into the original commit, resolving.

PR #450, opened 2024-04-01
(stdlib) `BigInt::pow`: Detect negative exponents and throw an exception

perlun:
This was detected while working on the `EvalHelper` methods to try and make more of them work in compiled mode too. This would be a noble goal, but for now feels kind of non-trivial. Because of this, the change likely does unfortunately not have excessive test coverage at the moment; we'll just have to take the compiler's word (Clang) for it that it works. :-)

### DIFF
--- a/src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
+++ b/src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
@@ -573,6 +573,11 @@ public class PerlangCompiler : Expr.IVisitor<object?>, Stmt.IVisitor<VoidObject>
 
                     // Probably enabled by -Wconversion, but this causes issues with valid code like `2147483647 /
                     // 4294967295U`.
+                    //
+                    // NOTE: disabling this also enables extremely dangerous constructs like being able to call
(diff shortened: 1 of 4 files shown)


src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
```diff
@@ -573,6 +573,11 @@ private void RegisterGlobalClasses()
 
                     // Probably enabled by -Wconversion, but this causes issues with valid code like `2147483647 /
                     // 4294967295U`.
+                    //
+                    // NOTE: disabling this also enables extremely dangerous constructs like being able to call
+                    // BigInt::pow(-3), causing the negative integer to be reinterpreted as a very large uint32 instead.
+                    // We should re-think this and really try to re-enable this, to prevent such horrible code from
+                    // compiling.
```
perlun, line 580: :point_up: The above was when `exponent` was a `uint32_t`. I tried to capture here some of the dangers of our current semantics. Not intending to do anything about this at the moment, just writing it down for future remembrance.

PR #451, opened 2024-04-02
(stdlib) Implement initial native `AsciiString` class

perlun:
"Native" as in "C++", being possible to access from C++ code targetting native machines.

### DIFF
--- a/src/Perlang.Tests.Integration/Typing/StringTests.cs
+++ b/src/Perlang.Tests.Integration/Typing/StringTests.cs
@@ -7,9 +7,11 @@ using static EvalHelper;
 
 public class StringTests
 {
-    [Fact]
+    [SkippableFact]
     public void string_variable_can_be_printed()
     {
+        Skip.If(PerlangMode.ExperimentalCompilation, "Not supported in compiled mode");
+
         string source = @"
                 var s: string = ""this is a string"";
 
@@ -22,9 +24,11 @@ public class StringTests
             .Be("this is a string");
     }
 
-    [Fact]
+    [SkippableFact]
     public void string_variable_can_be_reassigned()
     {
+        Skip.If(PerlangMode.ExperimentalCompilation, "Not supported in compiled mode");
+
         string source = @"
                 var s: string = ""this is a string"";
                 s = ""this is another string"";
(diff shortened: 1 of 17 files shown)


src/stdlib/src/utf8_string.h
```diff
@@ -0,0 +1,50 @@
...
+
+        // The backing byte array for this string. This is to be considered immutable and MUST NOT be modified at any
+        // point. There might be multiple UTF8String objects pointing to the same `bytes_`, so modifying one of them
+        // would unintentionally spread the modifications to these other objects too.
+        const char* bytes_;
+
+        // The length of the string in bytes, excluding the terminating `NUL` character.
+        size_t length_;
+    };
+}
```
perlun, line 50: I also threw in an `UTF8String` class while I was at it, and make them both derive from `perlang::String`. The bad part though: because of the way C++ works, making a `String` variable that you can freely assign and reassign content to is surprisingly hard. Will live with it for now.

The `UTF8String` obviously doesn't have any real "logic" at the moment, so to speak, but it exists so that we can properly represent non-ASCII strings in Perlang at least in some way.

src/Perlang.Tests.Integration/Typing/StringTests.cs
```diff
@@ -38,9 +38,16 @@ public void string_variable_can_be_reassigned()
...
 
-    [Fact]
+    [SkippableFact]
     public void ascii_string_inferred_variable_can_be_reassigned_with_non_ascii_value()
     {
+        // The code below is incredibly hard to support in compiled mode, because: an AsciiString cannot be assigned to a
+        // String variable in C++ (because the latter is an abstract class; I believe the C++ compiler will try to make a
+        // copy of it). `const perlang::string& s = ...` works, but then the problem is that the variable can obviously
+        // not be reassigned on the second line... because it is constant. We'll have to think through how to solve this
+        // properly.
```
perlun, line 48: https://github.com/perlang-org/perlang/pull/451/files#r1548516040

perlun:
Hmm, and should it be `AsciiString` or `ASCIIString`? We should at least be consistent (the other one is named `UTF8String` now :rofl:).

perlun:
> Hmm, and should it be AsciiString or ASCIIString? We should at least be consistent (the other one is named UTF8String now 🤣).

https://github.com/perlang-org/perlang/issues/370 has it as `AsciiString` and `Utf8String`, in fact. I'm not 100% certain about this.. but let's go with that for starters. Will rename `UTF8String` now.

(The other option would be to go the Golang route, and prefer that initialisms should be written in "consistent case", https://go.dev/wiki/CodeReviewComments#initialisms. I.e. `ASCIIString` and `UTF8String`. I'm probably a bit influenced by having read a bit of C# code in my days (and currently back to working with in my day-time job), but I kind of thing that `AsciiString` is somewhat kinder to the reader than `ASCIIString`. Yes, going with .NET-style for now, despite the "bad" part about in "changing case of existing acronyms".)

perlun:
Then again, `UTF8String` is kind of better than `Utf8String`.. :see_no_evil: Either way we put it.. we'll end up somewhat miserable. :neutral_face: Will pick one style now to just try to get this merged.

src/Perlang.Tests.Integration/Typing/TypingTests.cs
```diff
@@ -394,9 +413,11 @@ public void function_return_type_can_specify_explicit_type()
             Assert.Equal("typed return value", output);
         }
 
-        [Fact]
+        [SkippableFact]
         public void function_return_value_can_be_assigned_to_variable_of_same_type()
         {
+            Skip.If(PerlangMode.ExperimentalCompilation, "Not supported in compiled mode");
+
```
perlun, line 420: These exclusions are sad, but is because (non-pointer) `AsciiString` cannot be assigned to `String` (because of C++ copy semantics). I have an idea on how to workaround it though, with a bit of creative hacking... Will give it a try.

perlun: https://github.com/perlang-org/perlang/pull/453

Issue #454, opened 2024-04-11
Rewrite the Perlang compiler in Perlang


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/454. The GitHub page you are currently reading will not contain the latest information on this issue.

---

In other words, "let's become self-hosted". :slightly_smiling_face: 

With the effort done in #406, this is no longer a distant pipe dream. Here's the rough plan, as it currently stands:

- Start rewriting some of the simpler parts of the compiler in Perlang, perhaps beginning with a simple method or so:
  - Write it in Perlang
  - Add a mechanism in the Perlang compiler for compiling a given `.per` file to `.cc` and/or `.o` file, without producing a full executable. We could even add a `--lib` flag or something to indicate that we want to generate a library, but aggregating multiple `.per` files into a single `.cc` file if that's simplest for now could also be fine (note, will require generating the method definitions at the head of the file because of the way C++ works).
  - Use the existing Perlang compiler snapshot to compile its code to C++ (note: we'll probably need to "lock" the Perlang snapshot used here to a specific Perlang revision, to avoid surprises)
    - Given the "end goal" described below, we could even skip the "Use the existing Perlang compiler to compile its code"; we could just "build the code" and use that to regenerate the C++ parts, couldn't we? That way, there'll be no annoying binary snapshots to take care of etc.
  - Add the generated C++ code to the repo (in other words, commit the transpiled code as well for now)
  - Find a way to expose this C++ code to C#, using P/Invoke
  - Iterate on this for a while, rewriting more and more "simple" parts in Perlang. Try to find some low-hanging fruit to begin with, to get a general feeling of the feasibility of this.

### Challenges

Some parts will be reasonable to rewrite, like the ones described above, but other parts of the compiler will provide a significantly more challenging experience. I am thinking about the scanning, parsing and tree-walking parts here. Those are all "glued together" because of the data structures they use (C# classes to represent the different expression and statement types). We can either:

1. Do it as a "big-bang" conversion, where we convert _all relevant code_ to Perlang in one big PR. Not a great idea, and it will be a _huge_ undertaking to try to do it like this. And because of the immaturity of the language as it stands, we might even be sidetracked during the process by wanting/needing to add more language functionality to support the effort, delaying it even further.
2. Find a way to "inter-op" between C# and Perlang.

The latter seems "more doable" in a way, but then again:

- It's not trivial either, and
[...]

PR #455, opened 2024-04-11
(compiler) Add `-c` flag for "compile and assemble only"

perlun:
This commit is a small step towards #454. By adding this flag, it's possible to "use Perlang as a transpiler" only. Okay, a _little_ more than a transpiler, since it also emits ELF object files. :) But still, it helps us towards a state where we can gradually, incrementally rewrite parts of the Perlang compiler in Perlang. Oh, the joy! Oh, the glory!

### DIFF
--- a/src/Perlang.ConsoleApp/Program.cs
+++ b/src/Perlang.ConsoleApp/Program.cs
@@ -121,6 +121,7 @@ namespace Perlang.ConsoleApp
         {
             var versionOption = new Option<bool>(new[] { "--version", "-v" }, "Show version information");
             var detailedVersionOption = new Option<bool>("-V", "Show detailed version information");
+            var compileAndAssembleOnlyOption = new Option<bool>("-c", "Compile and assemble to a .o file, but do not produce and execute an executable");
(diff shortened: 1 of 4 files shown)


src/Perlang.ConsoleApp/Program.cs
```diff
@@ -200,9 +202,32 @@ public static int MainWithCustomConsole(string[] args, IPerlangConsole console)
...
+                        string scriptName = parseResult.GetValueForArgument(scriptNameArgument);
+
+                        // Console.Out.WriteLine(parseResult.Tokens);
+                        // Console.Out.WriteLine(parseResult.GetValueForArgument(scriptArguments));
+                        //
+                        // if (parseResult.Tokens.Count != 1)
+                        // {
+                        //     Console.Error.WriteLine("ERROR: When the -c option is used, no script arguments can be provided since the Perlang script/program will not get executed");
+                        //     return Task.FromResult(1);
+                        // }
```
perlun, line 219: Argh, did I just (well, yesterday) merge in commented-out code? :see_no_evil: Will go away in #456 for sure.

PR #456, opened 2024-04-12
(consoleapp) Removed custom `IConsole` implementations

perlun:
This is a preparation for rewriting the argument handling in Perlang, as one of the very first elements of the compiler to become self-hosted. Regretfully, this means that we will lose some of our test coverage, but we will have to revisit that aspect later.


src/Perlang.ConsoleApp/Program.cs
```diff
@@ -26,13 +25,6 @@ namespace Perlang.ConsoleApp
 {
     public class Program
     {
-        private static readonly ISet<string> ReplCommands = new HashSet<string>(new List<string>
-        {
-            "exit",
-            "help",
-            "quit"
-        }.Select(s => "/" + s));
-
```
perlun, line 35: No longer used, should rightfully have been removed already in #446.

src/Perlang.Tests/ConsoleApp/ProgramTest.cs
```diff
@@ -1,481 +0,0 @@
...
-
-                // This test ensures that the 'print 10' never gets executed. When a compiler _error_ occurs, program
-                // execution should be aborted.
-                StdoutContent.Should().NotContain("10");
-
-                StderrContent.Should().BeEmpty();
-            }
-        }
-    }
-}
```
perlun, line 481: Losing this whole test is the price we have to pay for removing the `IPerlangConsole` and related code. :cry: We'll have to live with it for now.

PR #459, opened 2024-04-14
(interpreter) Use `VisitorBase` everywhere applicable

perlun:
This is a preparation for other changes coming, where we add a new `IVisitor<Stmt>` method. Doing this is easier if `VisitorBase` is used everywhere possible, since it means we only have to add the implementation to the base class.

### DIFF
--- a/src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
+++ b/src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
@@ -51,6 +51,9 @@ namespace Perlang.Interpreter.Compiler;
 /// This is not the case for this simple implementation; it simply knows the level of indentation at each stage in the
 /// processing and does its best to try to produce something which is "not garbage", but not necessarily perfectly
 /// aesthetically pleasing.
+/// <remarks>Note: must **NOT** be a <see cref="VisitorBase"/>, because we use different return types than the base class'
+/// methods. Using the `new` keyword for these overloads will not work, since the base class will then not call the
+/// methods in this class as expected.</remarks>
 /// </summary>
 public class PerlangCompiler : Expr.IVisitor<object?>, Stmt.IVisitor<VoidObject>
 {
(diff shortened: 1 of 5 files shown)


src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
```diff
@@ -1467,7 +1468,7 @@ public VoidObject VisitVarStmt(Stmt.Var stmt)
         return VoidObject.Void;
     }
 
-    public VoidObject VisitWhileStmt(Stmt.While whileStmt)
+    public new VoidObject VisitWhileStmt(Stmt.While whileStmt)
```
perlun, line 1471: Hmm, this turned out to be more complex than I thought. :see_no_evil: I had to mark the new methods explicitly with `new` or `override` to get rid of the compiler warnings in this case.

I chose a strategy like this:

* Use `override` when possible (since that makes it possible to use the base class and still call the "right" method).
* Use `new` when we must, i.e. when we are changing the return type from the base class.

"WIll this even work?!?" when we don't properly `override` the base class? I _think_ so in fact... because we never really "use" the base class like this; we use the interface instead, in the polymorphic code.

---

This is indeed a bit of a mess, and serves as reasonable example of why the "OOP is bad" proponents _have_ a certain bit of a point. Don't worry, I will not throw out the baby with the bathwater and Perlang is likely to be an OOP language (or rather, "support OOP"). Still, it's a good case study for when OOP can get unpleasantly complex. Will create an issue about this and link/include this comment text there for reference.

perlun: Hmm, maybe it didn't work in fact. :thinking: https://github.com/perlang-org/perlang/actions/runs/8678694877/job/23795983071?pr=459 indicates this. That's probably because the _base class calls the base class methods_... which will go the wrong way for the `new` methods => the child class' methods will not get called, breaking the functionality.

So, in other words, "if you use OOP, you need a good test suite" might be another outcome of this. :joy:

perlun: Likewise with `PerlangInterpreter`; it had similar traits. Let's see if it compiles now. :tada:

Issue #457, opened 2024-04-13
Support inline C++


perlun:
Now that we are beginning the work on rewriting the Perlang compiler in Perlang (#454), we might find need for something interesting: _inline C++_.

Many C and C++ compilers support "inline assembly", i.e. being able to write certain methods in assembly language (and also take in/out parameters to these `asm` sections). We don't need that, at least not right now, but I think supporting similar for "inline C++" could be quite useful. One specific use case is when we will be called from C# code (using P/Invoke, i.e. `[DllImport(...)]`). In those cases, we might want to be able to e.g. receive a `const char** args` parameters (for easy marshalling of a `string[]` array on the C# side). However, this is currently not supported in Perlang and I really don't want to add support for unsafe, raw C-style strings (again, at least not right now). Letting the user specify sections of the program in C++ seems like a better and cleaner approach.

### Inline C++ in a Perlang function

```perlang
print "This is Perlang code";

c++ {
  // Raw C++ code here
  std::cout << "This is C++ code";
}
```

(The syntax for the above might be revised to be something like `cpp {}` instead, or even something like this:

```
`c++` {}
```

The problem with using something like `c++` is that it would cause an undesired ambiguity with a normal `c++;` postfix increment operator for a variable named `c`. This would destroy the language too much to be a real option.

### C++ prototypes (i.e. definitions)

This would be useful for defining the function prototypes for methods implemented in C++, for example if they need to be called from Perlang methods.

```perlang
#c++-prototypes

static do_something_useful(int i);

#/c++prototypes
```

### C++ declarations

This allows you to define one or more functions/methods in C++:

```
#c++-methods

extern "C" native_main(const char** args) 
{
    // ...
}

#/c++methods
```

perlun:
I think we're fine with this as of #462, closing away.

PR #462, opened 2024-04-16
(compiler) Add partial inline C++ support

perlun:
This change adds support for writing entire methods in C++. It lets you declare the prototype(s), using a special preprocessor directive, and similarly for the actual method implemention.

What we don't support is something like:

```
cpp {
  // C++ code here inside a Perlang function
}
```

...and this is also something we don't see a strong need for right now. Will leave this as-is, and might consider implementing the above at some point if the need arises.

### Related issue

* #457

### DIFF
--- a/src/Perlang.Common/Compiler/CompilerFlags.cs
+++ b/src/Perlang.Common/Compiler/CompilerFlags.cs
@@ -15,5 +15,11 @@ public enum CompilerFlags
     /// This flag disables all caching of compiled Perlang code, to ensure that all parts of the compilation is being
     /// used.
     /// </summary>
-    CacheDisabled = 1
+    CacheDisabled = 1,
+
+    /// <summary>
+    /// This flag which makes the compiler remove the `main` method if it is empty. This
+    /// can be useful e.g. when the `main` method is implemented in C++.
+    /// </summary>
+    RemoveEmptyMainMethod = 2
 }
--- a/src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
+++ b/src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
@@ -379,10 +379,14 @@ public class PerlangCompiler : Expr.IVisitor<object?>, Stmt.IVisitor<VoidObject>
         }
 
         ImmutableList<Stmt> statements;
+        ImmutableList<Token> cppPrototypes;
+        ImmutableList<Token> cppMethods;
 
         if (result.HasStatements)
         {
             statements = result.Statements!.ToImmutableList();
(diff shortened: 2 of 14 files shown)


src/Perlang.Common/Compiler/CompilerFlags.cs
```diff
@@ -15,5 +15,11 @@ public enum CompilerFlags
...
     /// used.
     /// </summary>
-    CacheDisabled = 1
+    CacheDisabled = 1,
+
+    /// <summary>
+    /// This flag which makes the compiler remove the `main` method if it is empty. This
+    /// can be useful e.g. when the `main` method is implemented in C++.
+    /// </summary>
+    RemoveEmptyMainMethod = 2
```
perlun, line 24: This was necessary to be able to write a Perlang program which is "bootstrapped" by a C++ `main` method implementation. :slightly_smiling_face: It's not an incredibly important feature, but it was useful to get a simple test in place without all the mechanisms for interfacing between Perlang and C++ code in place.

src/Perlang.Common/Stmt.cs
```diff
@@ -212,5 +222,29 @@ public override TR Accept<TR>(IVisitor<TR> visitor)
...
+                Token = token;
+                Content = content;
+                Type = type;
+            }
+
+            public override TR Accept<TR>(IVisitor<TR> visitor)
+            {
+                return visitor.VisitPreprocessorDirective(this);
+            }
+        }
```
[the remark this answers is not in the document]
perlun: I think I managed to solve this nicely now. By moving the "preprocessor" directives to the scanning phase instead, we can "slurp" the whole content between the opening and closing directive before it's even split up into tokens.

If/when we need some preprocessor stuff which is more parsing-oriented, like perhaps.. macros (Rust-style?), we might want to make these be some form of special statements or so, if we want to be able to visit them at the right place when tree-walking.. but it's very possible that we'll be able to handle such things at an early stage too. Time will tell.

src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
```diff
@@ -1481,6 +1519,23 @@ public VoidObject VisitWhileStmt(Stmt.While whileStmt)
...
+    //     }
+    //     else {
+    //         throw new PerlangCompilerException(
+    //             $"Unsupported preprocessor directive {preprocessorDirective.Type} encountered"
+    //         );
+    //     }
+    //
+    //     return VoidObject.Void;
+    // }
+
```
perlun, line 1538: Oops... :grin: Will be fixed in #473.

PR #463, opened 2024-04-23
(tests) Implement more `EvalHelper` methods for compiled mode

perlun:
This is a preparation for removing interpreted mode completely, expected to follow shortly.


perlun:
Failed in CI as expected, because we need the fix in #464 to make this work.

PR #466, opened 2024-04-27
(stdlib) Support ASCII-to-UTF8 string reassignment in compiled mode

perlun:


src/Perlang.Tests.Integration/Operator/Binary/DivisionTests.cs
```diff
@@ -132,7 +132,7 @@ public void division_by_zero_throws_expected_runtime_error()
         [SkippableFact]
         public void division_by_zero_halts_execution()
         {
-            Skip.If(PerlangMode.ExperimentalCompilation, "Not supported in compiled mode");
+            Skip.If(PerlangMode.ExperimentalCompilation, "Division by zero has undefined behavior in compiled mode.");
```
perlun, line 135: (Just a drive-by-improvement to make the skip message a bit clearer. We should try to do similarly on all skipped tests, ideally (or better yet, fix them so they are no longer skipped))

src/Perlang.Tests.Integration/Typing/StringTests.cs
```diff
@@ -42,16 +38,9 @@ public void string_variable_can_be_reassigned()
...
+    [Fact]
     public void ascii_string_inferred_variable_can_be_reassigned_with_non_ascii_value()
     {
-        // The code below is incredibly hard to support in compiled mode, because: an AsciiString cannot be assigned to a
-        // String variable in C++ (because the latter is an abstract class; I believe the C++ compiler will try to make a
-        // copy of it). `const perlang::string& s = ...` works, but then the problem is that the variable can obviously
-        // not be reassigned on the second line... because it is constant. We'll have to think through how to solve this
-        // properly.
-        Skip.If(PerlangMode.ExperimentalCompilation, "Not yet supported in compiled mode");
-
```
perlun, line 54: :point_up: This is the only test that didn't work in this test class. The other ones were just a matter of removing the skipping; they would already pass with our current code base.

src/Perlang.Tests.Integration/Typing/StringTests.cs
```diff
@@ -65,6 +54,23 @@ public void ascii_string_inferred_variable_can_be_reassigned_with_non_ascii_valu
...
+
+                print(s);
+            ";
+
+        var output = EvalReturningOutputString(source);
+
+        output.Should()
+            .Be("this is a string");
+    }
+
```
perlun, line 73: Added for the sake of completeness. `ASCIIString` was already doing the right thing, since its `from_static_string()` was returned a `std::shared_ptr<const ASCIIString>` return value.

src/stdlib/src/utf8_string.cc
```diff
@@ -1,21 +1,23 @@
 #include <cstring>
 #include <stdexcept>
+#include <memory>
 
 #include "utf8_string.h"
 
 namespace perlang
 {
-    UTF8String UTF8String::from_static_string(const char* s)
+    std::shared_ptr<const UTF8String> UTF8String::from_static_string(const char* s)
```
perlun, line 9: :point_up: This is the fix that unbreaks the test.

PR #470, opened 2024-05-04
(many) Support string concatenation in compiled mode

perlun:
Because we now have an idea around how to allocate Perlang strings dynamically (#453), we can also implement the `+` operator now which fixes one of the remaining "not supported in compiled mode" features.

### DIFF
--- a/src/Perlang.Common/ITypeReference.cs
+++ b/src/Perlang.Common/ITypeReference.cs
@@ -79,6 +79,7 @@ namespace Perlang
 
                 // Cannot use typeof(AsciiString) since Perlang.Common cannot depend on Perlang.Stdlib
                 var t when t.FullName == "Perlang.Lang.AsciiString" => true,
+                var t when t.FullName == "Perlang.Lang.Utf8String" => true,
                 var t when t.FullName == "Perlang.Lang.String" => true,
 
                 _ => false
--- a/src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
+++ b/src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
@@ -897,12 +897,22 @@ public class PerlangCompiler : Expr.IVisitor<object?>, Stmt.IVisitor<VoidObject>
                 // to ensure we get the expected operator on the C++ side (example: `int + uint` should produce a
                 // `long`, and at least one of the operands need to be widened to ensure expected semantics for an
                 // expression like `2 + 4294967295`)
-                if (expr.TypeReference.ClrType != expr.Left.TypeReference.ClrType)
+                //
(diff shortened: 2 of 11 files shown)


src/Perlang.Common/ITypeReference.cs
```diff
@@ -79,6 +79,7 @@ ClrType switch
 
                 // Cannot use typeof(AsciiString) since Perlang.Common cannot depend on Perlang.Stdlib
                 var t when t.FullName == "Perlang.Lang.AsciiString" => true,
+                var t when t.FullName == "Perlang.Lang.Utf8String" => true,
```
perlun, line 82: This was necessary for one of the tests to pass, and should be added here anyway.

src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
```diff
@@ -897,12 +897,22 @@ public object VisitAssignExpr(Expr.Assign expr)
...
+                        expr.TypeReference.ClrType == typeof(Lang.Utf8String)))
                 {
                     leftCast = expr.TypeReference.CppTypeCast;
                 }
 
-                if (expr.TypeReference.ClrType != expr.Right.TypeReference.ClrType)
+                if (expr.TypeReference.ClrType != expr.Right.TypeReference.ClrType && !(
+                        expr.TypeReference.ClrType == typeof(Lang.String) ||
+                        expr.TypeReference.ClrType == typeof(Lang.AsciiString) ||
+                        expr.TypeReference.ClrType == typeof(Lang.Utf8String)))
```
perlun, line 915: These were added here to avoid peculiar constructs like `*s1 + *(perlang::String)perlang::ASCIIString::from_static_string(" ")) + *s2` in the generated code. The problem is that while this would be kind-of-valid in C# and Java, we can't really cast to a `perlang::String` like this because this triggers an instantiation => `perlang::String` is abstract and cannot (and should not!) ever be instantiated. Ignoring the cast for now is the simplest approach here; let's see if all the tests are still passing or if this breaks something else now...

src/Perlang.Tests.Integration/Typing/StringTests.cs
```diff
@@ -41,44 +41,215 @@ public void string_variable_can_be_reassigned()
...
+
+                print(s2);
+                """;
+
+        var output = EvalReturningOutputString(source);
+
+        output.Should()
+            .Be("Årets varmaste temperatur (Celsius): 32");
+    }
+
```
perlun, line 244: These tests were added while working on this, but they are not yet operational because the underlying (C++) stdlib code isn't there yet. Will be fixed in an upcoming PR.

PR #473, opened 2024-05-07
(many) Support more types in string+int and int+string concatenation

perlun:


src/Perlang.Interpreter/Typing/TypeResolver.cs
```diff
@@ -127,36 +127,36 @@ public override VoidObject VisitBinaryExpr(Expr.Binary expr)
...
                     else if (expr.Operator.Type == TokenType.PLUS &&
-                             (leftTypeReference.ClrType == typeof(int) &&
+                             (new[] { typeof(int), typeof(long), typeof(uint), typeof(ulong) }.Contains(typeof(int)) &&
                               rightTypeReference.ClrType == typeof(AsciiString))) {
                         // 42 + "string" + 42
                         expr.TypeReference.ClrType = typeof(AsciiString);
                     }
                     else if (expr.Operator.Type == TokenType.PLUS &&
-                             (leftTypeReference.ClrType == typeof(int) &&
+                             (new[] { typeof(int), typeof(long), typeof(uint), typeof(ulong) }.Contains(typeof(int)) &&
```
perlun, line 159: Hmm, these three are quite obviously wrong. :grin: Let's see if we have a test that covers this error of mine... :see_no_evil:

perlun: I think we do, in fact. The `[Internal compiler error`](error: no viable conversion from 'shared_ptr)` problem is _probably_ caused by this, triggering a different code path, making a previously skipped test no longer be skipped... and that test in turn is currently broken, since it generates C++ code which looks like this:

```c++
std::shared_ptr<const perlang::ASCIIString> a = perlang::ASCIIString::from_static_string("a");
perlang::print(a);
std::shared_ptr<const perlang::ASCIIString> b = (*a + *perlang::ASCIIString::from_static_string(" b"));
perlang::print(b);
std::shared_ptr<const perlang::ASCIIString> c = (*a + *perlang::ASCIIString::from_static_string(" c"));
perlang::print(c);
std::shared_ptr<const perlang::ASCIIString> d = (*b + *perlang::ASCIIString::from_static_string(" d"));
perlang::print(d);
```

This fails because some of these `+` operators return a `std::shared_ptr<const perlang::String>`, which can't be converted to `std::shared_ptr<const perlang::ASCIIString>`. We'll have to look into this at some point, but not tonight...

PR #474, opened 2024-05-09
(many) Fix `ASCIIString+ASCIIString` concatenation to return `ASCIIString`

perlun:
The problem with letting it return `String` (as was the case before this fix) is that the result couldn't be assigned to an `ASCIIString` variable, but the `TypeResolver` would still believe the result of such an operation would be a pointer to `ASCIIString`. The end result would be that we would emit uncompilable C++ code.

### DIFF
--- a/src/Perlang.Interpreter/Typing/TypeResolver.cs
+++ b/src/Perlang.Interpreter/Typing/TypeResolver.cs
@@ -171,7 +171,7 @@ namespace Perlang.Interpreter.Typing
                              (leftTypeReference.ClrType == typeof(AsciiString) &&
                               rightTypeReference.ClrType == typeof(AsciiString))) {
                         // "string" + "string"
-                        expr.TypeReference.ClrType = typeof(Utf8String);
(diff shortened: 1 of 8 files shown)


src/Perlang.Interpreter/Typing/TypeResolver.cs
```diff
@@ -171,7 +171,7 @@ public override VoidObject VisitBinaryExpr(Expr.Binary expr)
                              (leftTypeReference.ClrType == typeof(AsciiString) &&
                               rightTypeReference.ClrType == typeof(AsciiString))) {
                         // "string" + "string"
-                        expr.TypeReference.ClrType = typeof(Utf8String);
+                        expr.TypeReference.ClrType = typeof(AsciiString);
```
perlun, line 174: :point_up: This was a bug; `ASCIIString+ASCIIString` should obviously not return a `UTF8String`. However, both of these would fail equally because of the missing supporting code in the C++ stdlib (fixed in this PR).

PR #475, opened 2024-05-09
(many) Support string+bigint and bigint+string concatenation

perlun:


src/stdlib/src/ascii_string.cc
```diff
@@ -39,7 +59,7 @@ namespace perlang
     ASCIIString::~ASCIIString()
     {
         if (owned_) {
-            delete[] bytes_;
+            free((void*)bytes_);
```
perlun, line 62: This was an oversight in the old code; it _worked_ but.. mixing `malloc` and `delete` isn't really a great idea. Either go with `malloc`+`free` or `new`+`delete` but never a combination of those. Since we are dealing with plain, C-style `char*` pointers that we then do `memcpy` and stuff using, I think C-style allocation makes a bit of sense in this case.

src/stdlib/src/ascii_string.cc
```diff
@@ -157,8 +192,8 @@ namespace perlang
         size_t length = str.length() + rhs.length();
         char *bytes = new char[length + 1];
 
-        memcpy((void*)bytes, str.c_str(), str.length());
-        memcpy((void*)(bytes + str.length()), rhs.bytes(), rhs.length());
+        memcpy(bytes, str.c_str(), str.length());
+        memcpy((bytes + str.length()), rhs.bytes(), rhs.length());
```
perlun, line 196: (Removed these `void*` casts both here and in `utf8_string.cc`. `char*` can be implicitly cast to `void*`, so let's perhaps use that fact to make the code a little bit more readable.

PR #478, opened 2024-05-11
(test) Enable string bounds checking test

perlun:
The code already seemeed to be there since
https://github.com/perlang-org/perlang/pull/451, so let's reap the full benefit of it now and adjust the code accordingly. Also added a similar `UTF8String` test for completeness.

### DIFF
--- a/src/Perlang.Tests.Integration/IndexOperator/StringIndexing.cs
+++ b/src/Perlang.Tests.Integration/IndexOperator/StringIndexing.cs
@@ -7,44 +7,61 @@ namespace Perlang.Tests.Integration.IndexOperator;
 public class StringIndexing
 {
     [Fact]
-    public void AsciiString_can_be_indexed_by_integer()
+    public void ASCIIString_can_be_indexed_by_integer()
     {
-        string source = @"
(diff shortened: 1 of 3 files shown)


src/Perlang.Tests.Integration/IndexOperator/StringIndexing.cs
```diff
@@ -7,44 +7,61 @@ namespace Perlang.Tests.Integration.IndexOperator;
...
+            .Message.Should().Contain("exited with exit code 134");
+
+        result.OutputAsString.Should()
+            .Contain("Index 10 is out-of-bounds for a string with length 6");
+    }
 
-        string source = @"
-            print ""foobar""[10];
-        ";
+    [Fact(Skip = "Not yet supported for UTF8String")]
```
perlun, line 39: I'll go ahead and fix this, but in a separate PR.

perlun: Ah, my bad. This shouldn't be "fixed" for the time being; it's not something we can easily support. #479

PR #488, opened 2024-06-06
(compiler) Add negative test for indexing string variables

perlun:


src/Perlang.Interpreter/Typing/TypeResolver.cs
```diff
@@ -498,19 +498,6 @@ public override VoidObject VisitIndexExpr(Expr.Index expr)
...
-                    {
-                        typeValidationErrorCallback(new TypeValidationError(
-                            expr.ClosingBracket,
-                            $"'string' cannot be indexed by '{argumentType.ToTypeKeyword()}'")
-                        );
-                    }
-
-                    expr.TypeReference.ClrType = typeof(char);
-                    break;
-
```
perlun, line 513: I was initially looking into fixing this, but I changed my mind. Remember that Perlang strings are one of the following:

* `ASCIIString`
* `UTF8String`

The former can easily be indexed, whereas it's non-obvious what indexing the second one would _mean_, semantically. I'll make a note about this in #370 because I'm actually getting rather fed up with UTF-8 strings limiting the `string` interface so much.

PR #490, opened 2024-06-10
(common) Make Initializer property be nullable

perlun:
It was already nullable in practice. This change adjusts the code to take this into account.

### DIFF
--- a/src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
+++ b/src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
@@ -1542,7 +1542,7 @@ public class PerlangCompiler : Expr.IVisitor<object?>, Stmt.IVisitor<VoidObject>
 
         currentMethod.Append($"{Indent(indentationLevel)}{stmt.TypeReference.PossiblyWrappedCppType} {variableName}");
 
-        if (stmt.HasInitializer)
+        if (stmt.Initializer != null)
         {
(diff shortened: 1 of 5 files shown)


src/Perlang.Interpreter/Typing/TypeResolver.cs
```diff
@@ -786,6 +786,7 @@ public override VoidObject VisitVarStmt(Stmt.Var stmt)
             return VoidObject.Void;
         }
 
+        // TODO: Figure out why this is unused
```
perlun, line 789: Seems to have been unused since de3ff2ef925b6244dea54d39978bdbe221df8af7.

perlun: Fixed in #491, closing.

src/Perlang.Interpreter/Compiler/PerlangCompiler.cs
```diff
@@ -1542,7 +1542,7 @@ public VoidObject VisitVarStmt(Stmt.Var stmt)
 
         currentMethod.Append($"{Indent(indentationLevel)}{stmt.TypeReference.PossiblyWrappedCppType} {variableName}");
 
-        if (stmt.HasInitializer)
+        if (stmt.Initializer != null)
```
perlun, line 1545: This kind of change :point_up: is actually more than just "removing a property getter". It's actually _required_ now that `Initializer` is nullable, to avoid CS8062 nullability warnings.

Issue #487, opened 2024-06-01
Add Perlang array support


perlun:
# Moved to GitLab

Please continue to the new version of this issue here: https://gitlab.perlang.org/perlang/perlang/-/issues/487. The GitHub page you are currently reading will not contain the latest information on this issue.

---

Here's a quote from myself in a commit message for what I just planned to start working on (yes, I sometimes write the commit message before writing the implementation. :slightly_smiling_face:)

> Perlang arrays are implemented as instances of our C++ `PerlangArray`
> (non-template) type. The generic part of it is implemented on the
> Perlang side; the actual arrays store essentially a fixed-size C array
> of `void *` pointers which are cast to the appropriate type in the
> generated C++ code. We also special-case certain types like `IntArray`,
> `LongArray` and so forth, for the value types currently supported in
> Perlang.

I might not implement all of this in a single PR/branch, so I'm copying it here to keep it as some form of "design notes" for this feature.

Also, we might be able to deviate a bit from the above; we can even add `StringArray` to make it fully "specific" for now. Since there's not really any way to declare user-defined types at the moment (neither classes nor structs), we only need to support arrays for the Perlang-predefined types since that's all you'll ever going to need as a Perlang developer right now.

perlun:
Has been started in #494 and #496 (i.e. milestone 0.5.0), but the feature is not really "complete" in that sense. Marking as 0.6.0 for clarity.

PR #494, opened 2024-06-12
(many) Add implicitly typed `int` arrays

perlun:
This is an important step towards completing #487, but a few more things are needed for this:

* Being able to specify `int[]` as an explicit type specifier. Right now, only implicitly typed `int` arrays are supported.
* Being able to use this for other data types like `long[]`, `double[]` and `string[]`.

As for the underlying `perlang::IntArray` class, what I am aiming for here is something a bit more high-level than traditional C-style `int[]` arrays. They are nice in the sense of being "low overhead" but are also prone to all kinds of C-style memory issues (heap overflow, most notably). Perlang arrays will be more akin to Java or C#-style arrays, which are inherently "safe by design". It should be more or less impossible to use a Perlang array in such a way as to cause the program to crash. Because we don't have a JVM or MSIL interpreter to help us with this, we have to implement the boundary checks on our own.

At the moment, those boundary checks throw (C++) exceptions. We could consider adding more "try"-oriented methods too, like `try_get(index, default_value)` or even `try_get(index, () -> default_value)` if we want to be more fancy. Either way, the point is to avoid exceptions and let the caller (by returning a `bool`) check if the operation was successful or not by different, sometimes more appropriate means than exceptions.


src/Perlang.Tests.Integration/Arrays/IntArrayTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        output.Should()
+            .BeEquivalentTo(
+                "1",
+                "2",
+                "3"
+            );
+    }
+
+    // TODO: Add test for indexing before and after array size, ensuring that we get the expected exceptions
```
perlun, line 51: Hmm, would have been nice to take in the same PR I guess. :see_no_evil: Will have to come in a followup PR.

perlun: https://github.com/perlang-org/perlang/pull/496/files#r1644983171

PR #496, opened 2024-06-18
(many) Add explicitly typed `int[]` arrays

perlun:
Building on top of #494, but making it possible to declare these explicitly too.

### DIFF
--- a/src/Perlang.Interpreter/Typing/TypeResolver.cs
+++ b/src/Perlang.Interpreter/Typing/TypeResolver.cs
@@ -832,7 +832,7 @@ namespace Perlang.Interpreter.Typing
             return VoidObject.Void;
         }
 
-        private static void ResolveExplicitTypes(ITypeReference typeReference)
+        private void ResolveExplicitTypes(ITypeReference typeReference)
         {
             if (typeReference.TypeSpecifier == null)
             {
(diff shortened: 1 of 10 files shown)


src/Perlang.Interpreter/Typing/TypeResolver.cs
```diff
@@ -853,41 +853,48 @@ private static void ResolveExplicitTypes(ITypeReference typeReference)
...
 
                 case "double" or "Double":
-                    typeReference.SetClrType(typeof(double));
+                    typeReference.SetClrType(typeReference.IsArray ? typeof(double[]) : typeof(double));
                     break;
 
                 case "string" or "String":
                     // Perlang String is NOT the same as the .NET String class.
-                    typeReference.SetClrType(typeof(Lang.String));
+                    typeReference.SetClrType(typeReference.IsArray ? typeof(Lang.String[]) : typeof(Lang.String));
```
perlun, line 887: The above adds support for many more array types than just `int[]`, but only `int[]` (implicit and explicit) is supported right now. We'll add `int[]` and `string[]` as the two first ones, and the next ones will perhaps go into the next milestone.

src/Perlang.Tests.Integration/Arrays/IntArrayTests.cs
```diff
@@ -48,5 +48,44 @@ public void implicitly_typed_int_array_can_be_indexed()
...
+        var result = EvalWithRuntimeErrorCatch(source);
+
+        result.Errors.Should()
+            .ContainSingle()
+            .Which
+            .Message.Should().Contain("exited with exit code 134");
+
+        result.OutputAsString.Should()
+            .Contain("index out of range (3 > 2)");
+    }
```
perlun, line 90: https://github.com/perlang-org/perlang/pull/494/files#r1644933209

Issue #423, opened 2024-02-17
Move the Perlang project hosting to self-hosted GitLab instance


perlun:
**MOVED TO GITLAB: https://gitlab.perlang.org/perlang/perlang/-/issues/423**; the GitHub page you are currently reading will not contain the latest information on this issue.

---

At some point, I believe we should move the Perlang hosting to a self-hosted GitLab instance, for a few different reasons:

- **Better autonomy.** As the Debian project [describes it](https://github.com/debian):

  > This account is not intended to serve as the canonical (specified with Vcs-* fields of debian/control) location for corresponding Debian source packages. Most often such repositories should be made available on the Debian project's public forge Salsa to guarantee autonomy.

  "Autonomy" in the previous sentence links to this page: https://web.archive.org/web/20081002034726/http://autonomo.us/2008/07/franklin-street-statement/

  Using GitLab means we, as a project, will have full control over the repository. GitHub outage, because of problems with their hosting? Forget it. Of course, we can instead have "our own" outages, but still, this give us more control over a critical part of our own infrastructure.

- **More freedom.** We are currently relying upon GitHub (which is a proprietary, SaaS-hosted software). The free (as in freedom) version of GitLab, namely [GitLab Community Edition](https://gitlab.com/gitlab-org/gitlab-foss/) is licensed under [the MIT Expat license](https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/LICENSE), a license which qualifies as Free Software by the FSF and Open Source by the OSI.

- **Be able to use GitLab CI.** GitHub Actions work quite well, but it does have certain disadvantages. One of the things I   particularly dislike is the "Status checks that are required" which is configured via "branch protection rules" under https://github.com/perlang-org/perlang/settings/branches. Here, the required status checks are configured. There's just one problem with this: it makes the CI settings _configured in the web interface_ be tightly coupled to the action names defined under [`.github/workflows`](https://github.com/perlang-org/perlang/tree/master/.github/workflows). You rename one of these, and the CI settings break. I find this rather fragile and I think the GitLab CI approach - where much more of this is defined in [`.gitlab-ci.yml`](https://docs.gitlab.com/ee/ci/yaml/) is much preferable.

  - Using GitLab CI with GitHub [can be done](https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html) though, but:
    1. It requires GitLab Premium or Ultimate (i.e. the paid-for tiers)
[...]

PR #506, opened 2024-09-19
(ci) Fix two bugs in package building

perlun:
- The first bug was critical. The fact that `perlang_cli.so` wasn't being built meant that `perlang -v` would subsequently fail, failing the whole snapshot building and publishing.

- Equally important but more subtle in nature, the `perlang_cli.so` was also not properly included in the generated `tar.gz` files. This meant that the binaries would not work properly, even if `.tar.gz` files would have been generated.

### DIFF
--- a/.github/workflows/publish-snapshot-packages.yml
+++ b/.github/workflows/publish-snapshot-packages.yml
@@ -50,8 +50,16 @@ jobs:
       - name: Build host-compatible release
         run: dotnet build
 
+      # This is needed to be able to run `perlang -v` a few commands below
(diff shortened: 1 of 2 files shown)


perlun:
(#423, I hear you. :slightly_smiling_face: The reason this PR was done on GitHub after the move was because this is untestable without GitHub Actions. Once this is CI clean, I'll cherrypick the commit into our GitLab repo and close the PR)

perlun:
Merged in b2833056947e0131173c6ea3ff31c824ea417e31, closing PR.

